PR #2, opened 2021-02-02
chore: sample action client

tomer-epstein:


types/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@sap-devx/bas-platform-types",
   "publisher": "SAPOSS",
-  "version": "0.0.33",
+  "version": "0.0.13",
```
ira-gordin-sap, line 4: Is it by matter?

package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "bas-platform",
 	"displayName": "App-Studio-Toolkit Extension",
 	"description": "SAP Business Application Studio Toolkit extension",
-	"version": "0.0.33",
+	"version": "0.0.13",
```
ira-gordin-sap, line 5: Is it by matter?

ira-gordin-sap (approved):
I hope versions changed by matter.

types/package.json
tomer-epstein: Yes. Yes. Later we will decide both when to bump it to 1.0.0

package.json
tomer-epstein: Yes. Later we will decide both when to bump it to 1.0.0

PR #28, opened 2021-03-11
Adding id parameter to Action object

NetaMolcho:


package.json
```diff
@@ -42,6 +42,7 @@
 		"package": "vsce package ."
 	},
 	"dependencies": {
+		"@sap/consume-services-types": "^0.3.6",
```
idoprz, line 45: The toolkit should not be dependant on the consume services. It is lower level extension.
You should either define the IService in the toolkit, or use some more "abstract type" (preferable not any)

src/actions/interfaces.ts
```diff
@@ -1,4 +1,5 @@
 import * as vscode from "vscode";
+import { IService } from "@sap/consume-services-types";
```
idoprz, line 2: remove this dependency...

src/actions/interfaces.ts
```diff
@@ -17,10 +18,11 @@ export enum ActionJsonKey {
 
 export type CommandActionParams = any [];
 export type ExecuteActionParams = any [];
-export type SnippetActionParams = any;
+export type SnippetActionParams = any | {data: any, service: IService} | {data: any};
```
idoprz, line 21: it can be service: any
or better to define something in here that will cover the service type

idoprz (commented):
see my comments in line. otherwise it's fine by me, but Tomer should approve. not me.

idoprz (dismissed):
👍

tomer-epstein (changes requested):
increment version 0.0.4--> https://github.com/NetaMolcho/app-studio-toolkit/blob/idAction/package.json#L5
increment version 0.0.4 (too) --> https://github.com/NetaMolcho/app-studio-toolkit/blob/idAction/types/package.json#L4

tomer-epstein (approved):

PR #35, opened 2021-03-21
ci: bump version to 0.0.5

NetaMolcho:


tomer-epstein (changes requested):
No change smade to the types. why bump it ?

PR #36, opened 2021-03-22
fix: forEach on all of the actions

NetaMolcho:


src/actions/controller.ts
```diff
@@ -8,7 +8,10 @@ export class ActionsController {
     public static loadActions() {
       vscode.extensions.all.forEach((extension) => {
         if (extension?.packageJSON?.BASContributes?.actions) {
-          this.actions.push(extension.packageJSON.BASContributes.actions);
+          const newActions: IAction[] = extension?.packageJSON?.BASContributes?.actions;
+          newActions.forEach((action) => {
+            this.actions.push(action);
+          });
```
tomer-epstein, line 14: extension.packageJSON.BASContributes.actions.forEach((action) => {
            this.actions.push(action);
          });

tomer-epstein (changes requested):

src/actions/controller.ts
NetaMolcho: it needs to know that this is IAction[] type, it doesn't accept it without a declaration

tomer-epstein (changes requested):
increment app-studio-toolkit version to 0.0.6

tomer-epstein (dismissed):

src/actions/controller.ts
tomer-epstein: > it needs to know that this is IAction[] type, it doesn't accept it without a declaration

(extension.packageJSON.BASContributes.actions as IAction[]).forEach

src/actions/controller.ts
```diff
@@ -8,7 +8,9 @@ export class ActionsController {
     public static loadActions() {
       vscode.extensions.all.forEach((extension) => {
         if (extension?.packageJSON?.BASContributes?.actions) {
-          this.actions.push(extension.packageJSON.BASContributes.actions);
+          (extension.packageJSON.BASContributes.actions as IAction[]).forEach((action) => {
+            this.actions.push(action);
+          });
         }
```
tomer-epstein, line 14: Can you please add a test for this method?

tests/api.spec.ts
```diff
@@ -5,6 +5,7 @@ import * as _ from "lodash";
 import * as sinon from "sinon";
 import { IAction, ActionType } from "../src/actions/interfaces";
 import { ActionsController } from '../src/actions/controller';
+import * as vscode from "vscode";
```
tomer-epstein, line 8: Please remove.
We are using --> mockVscode

tests/api.spec.ts
```diff
@@ -77,6 +79,30 @@ describe("api unit test", () => {
...
+                    }]
+                },
+            }
+        }]
+        _.set(vscode, "extensions.all", allExtensioms);
+
+        const action: IAction = {
+			"id" : "abc123",
+			"actionType" : ActionType.Command,
+		}
```
tomer-epstein, line 98: const action: IAction = {
			"id" : "abc123",
			"actionType" : ActionType.Command,
		}
        const allExtensioms = [{
            packageJSON: {
                "BASContributes": {
                    "actions": [action]
                },
            }
        }]
        _.set(vscode, "extensions.all", allExtensioms);

tomer-epstein (changes requested):

tests/api.spec.ts
NetaMolcho: you have the vscode import for this set

tests/api.spec.ts
NetaMolcho: it's impossible to do set on mockVscode

tests/api.spec.ts
NetaMolcho: allExtensions should resemble package.json, so it should contain strings

NetaMolcho: allExtension should resemble package.json, so it should only contain strings

tomer-epstein (approved):

PR #46, opened 2021-04-19
feat: add getParameter api

ira-gordin-sap:
It was tested manually and works. I need to write 2 more tests after I'll find how to mock the '@sap/plugin' module.
I'll really appreciate you can start a review meanwhile, because the end of the sprint is coming. 
Completed including manual and unit tests.

### DIFF
--- a/.mocharc.yaml
+++ b/.mocharc.yaml
@@ -1,0 +1,6 @@
+require: 
+  - 'ts-node/register/transpile-only'
+  - 'source-map-support/register'
+recursive: 
+  - 'tests/**/*.spec.ts'
+timeout: 80000
--- a/package.json
+++ b/package.json
@@ -62,8 +62,9 @@
 		"package": "vsce package ."
 	},
 	"dependencies": {
+		"@vscode-logging/wrapper": "^1.0.0",
 		"lodash": "4.17.21",
-		"@vscode-logging/wrapper": "^1.0.0"
+		"optional-require": "^1.0.3"
 	},
 	"devDependencies": {
 		"@commitlint/cli": "12.0.1",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
 	"name": "app-studio-toolkit",
 	"displayName": "App-Studio-Toolkit Extension",
 	"description": "SAP Business Application Studio Toolkit extension",
-	"version": "0.0.6",
+	"version": "0.0.7",
 	"publisher": "SAPOSS",
 	"license": "Apache-2.0",
 	"icon": "icon.png",
@@ -89,7 +89,8 @@
 		"vsce": "^1.69.0",
 		"vscode-test": "^1.2.2",
 		"coveralls": "2.11.16",
-		"@vscode-logging/types": "0.1.3"
+		"@vscode-logging/types": "0.1.3",
+		"mock-require": "^3.0.3"
 	},
 	"husky": {
 		"hooks": {
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
 		"compile": "tsc -p ./",
 		"watch": "tsc -watch -p ./",
 		"pretest": "npm run compile",
-		"test": "nyc mocha -p tsconfig.json --recursive --no-parallel tests/**/*.spec.ts",
+		"test": "nyc mocha -p tsconfig.json --config ./.mocharc.yaml --no-parallel",
 		"lint": "eslint . --ext .ts,.tsx --cache",
 		"lint:fix": "eslint . --ext .ts,.tsx --cache --fix",
 		"package": "vsce package ."
--- a/sample-action-client/src/extension.ts
+++ b/sample-action-client/src/extension.ts
@@ -23,12 +23,12 @@ export async function activate(context: vscode.ExtensionContext) {
 
 	context.subscriptions.push(vscode.commands.registerCommand("get.parameter", async () => {
         const parameterName = await vscode.window.showInputBox({ prompt: "Enter Parameter Name", ignoreFocusOut: true });
-        if (!parameterName) { 
-            // undefined in escape and empty on enter
+        // undefined on escape and empty on enter
+        if (parameterName === undefined || parameterName === "") {
             return;
         }        
         
-        const parameterValue = await basAPI.getParameter(parameterName as string);
+        const parameterValue = await basAPI.getParameter(parameterName);
         vscode.window.showInformationMessage(`${parameterValue} returned for ${parameterName}`,);
     }));
 
--- a/sample-action-client/src/extension.ts
+++ b/sample-action-client/src/extension.ts
@@ -20,6 +20,18 @@ export async function activate(context: vscode.ExtensionContext) {
         vscode.window.showInformationMessage(`Action ${action.name} was scheduled to run`,);
         basAPI.actions.performAction(action, {schedule: true});
     }));
+
+	context.subscriptions.push(vscode.commands.registerCommand("get.parameter", async () => {
+        const parameterName = await vscode.window.showInputBox({ prompt: "Enter Parameter Name", ignoreFocusOut: true });
+        if (!parameterName) { 
+            // undefined in escape and empty on enter
+            return;
+        }        
+        
+        const parameterValue = await basAPI.getParameter(parameterName as string);
+        vscode.window.showInformationMessage(`${parameterValue} returned for ${parameterName}`,);
+    }));
+
 }
 
 export function deactivate() {
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,8 +1,8 @@
 import * as vscode from 'vscode';
-//import {isEmpty, get} from 'lodash';
 import { performAction } from './actions/client';
 import { ActionsController } from './actions/controller';
 import { ExecuteAction, SnippetAction, CommandAction, FileAction } from './actions/impl';
+import { getParameter } from './apis/parameters';
 import { getLogger } from './logger/logger';
 export * from "./actions/interfaces";
 
@@ -38,30 +38,7 @@ export const bas = {
     },
 
     async getParameter(parameterName: string) : Promise<string | undefined> {
-        // const onBasEnv = "env.WS_BASE_URL";
-        // const isInBAS = !isEmpty(get(process, onBasEnv));
-        //const isInBAS = get(process, onBasEnv)?.is;
-        // if (!isInBAS) {
-        //     getLogger().trace("Running in VS Code, so returning undefined.");
-        //     return undefined;
-        // }
-        // getLogger().trace("Running on BAS.");
-
-        const optionalRequire = require("optional-require")(require);
-        const noSapPlugin = "noSapPlugin";
-        const sapPlugin = optionalRequire('@sap/plugin') || noSapPlugin;
-        if (sapPlugin === noSapPlugin) {
-            getLogger().error("Failed to load @sap/plugin, so returning undefined.");
-            return undefined;
-        } 
-        getLogger().trace("@sap/plugin successfully loaded.");
-
-        const configuration = await sapPlugin.window.configuration();
-        getLogger().trace("Configuration successfully received.", {configuration});
-
-        const parameterValue = configuration?.[parameterName];
-        getLogger().trace(`configuration[${parameterName}]=${parameterValue}`, {parameterName, parameterValue});
-        return parameterValue;
+        return getParameter(parameterName);
     },
 
     actions: {
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,10 +1,10 @@
 import * as vscode from 'vscode';
+//import {isEmpty, get} from 'lodash';
 import { performAction } from './actions/client';
 import { ActionsController } from './actions/controller';
 import { ExecuteAction, SnippetAction, CommandAction, FileAction } from './actions/impl';
 import { getLogger } from './logger/logger';
 export * from "./actions/interfaces";
-import _ = require('lodash');
 
 export const bas = {
     getExtensionAPI: <T>(extensionId: string): Promise<T> => {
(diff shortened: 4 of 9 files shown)


bd82:
please don't do any more force pushes as it makes it harder to do incremental review

ira-gordin-sap:
sorry, I was sure you did not start - I only removed unused comment

sample-action-client/src/extension.ts
```diff
@@ -20,6 +20,18 @@ export async function activate(context: vscode.ExtensionContext) {
         vscode.window.showInformationMessage(`Action ${action.name} was scheduled to run`,);
         basAPI.actions.performAction(action, {schedule: true});
     }));
+
+	context.subscriptions.push(vscode.commands.registerCommand("get.parameter", async () => {
+        const parameterName = await vscode.window.showInputBox({ prompt: "Enter Parameter Name", ignoreFocusOut: true });
+        if (!parameterName) {
```
bd82, line 26: it is better to avoid automatic boolean conversions in JavaScript
as it can lead to strange bugs in edge cases.

Something like:
```javascript
// undefined on escape and empty on enter
if (parameterName === undefined || parameterName === "") { /* ... */} 
```

sample-action-client/src/extension.ts
```diff
@@ -20,6 +20,18 @@ export async function activate(context: vscode.ExtensionContext) {
...
     }));
+
+	context.subscriptions.push(vscode.commands.registerCommand("get.parameter", async () => {
+        const parameterName = await vscode.window.showInputBox({ prompt: "Enter Parameter Name", ignoreFocusOut: true });
+        if (!parameterName) { 
+            // undefined in escape and empty on enter
+            return;
+        }        
+        
+        const parameterValue = await basAPI.getParameter(parameterName as string);
```
bd82, line 31: I think the explicit casting here is not "at the correct spot".

Option1:
- add the casting where you defined `parameterName` (what is the return type of `showInputBox` ?)

Option2:
- extract the `if (!parameterName) {` block to a helper function with a **type guard** 
- https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

we can discuss which option is better offline

ira-gordin-sap: I'll remove the casting. I don't really need it now. Don't remember why I added it.

src/api.ts
```diff
@@ -2,7 +2,9 @@ import * as vscode from 'vscode';
 import { performAction } from './actions/client';
 import { ActionsController } from './actions/controller';
 import { ExecuteAction, SnippetAction, CommandAction, FileAction } from './actions/impl';
+import { getLogger } from './logger/logger';
 export * from "./actions/interfaces";
+import _ = require('lodash');
```
bd82, line 7: I recommend you only import the exact functions you need from lodash
You can use an ESM compatible lodash build from here: https://www.npmjs.com/package/lodash-es

bd82, line 7: import order:
- don't mix import and exports. (export after all import)
- import of 3rd party libraries should appear **before** import of relative paths modules.

tests/api.spec.ts
```diff
@@ -126,3 +175,12 @@ function promiseWithTimeout(promise: any, timeout: number){
...
 }
+
+/**
+ * Stub key-> value object to the environment variable for testing purpose
+ * @param newValues - The key-> valye to set the environment
+ */
+ function stubEnv(newValues: Record<string, unknown>): void {
+    const extendedEnv = extend({}, process.env, newValues);
+    stub(process, "env").value(extendedEnv);
+  }
```
bd82, line 186: missing newline

src/api.ts
```diff
@@ -35,6 +37,32 @@ export const bas = {
         return ActionsController.getAction(actionId);
     },
 
+    async getParameter(parameterName: string) : Promise<string | null> {
```
bd82, line 40: why is it `string | null` is it not more appropriate to return `string | undefined` 

WDYT?

src/api.ts
```diff
@@ -35,6 +37,32 @@ export const bas = {
...
+        const onBasEnv = "env.WS_BASE_URL";
```
bd82, line 41: `"env.WS_BASE_URL";` to constants instead of magic string value?

src/api.ts
ira-gordin-sap: import {isEmpty, get} from 'lodash';
?
(We have _ in all our projects..)

src/api.ts
bd82: ignore this, see next commt

src/api.ts
ira-gordin-sap: @idoprz asked it

src/api.ts
```diff
@@ -35,6 +37,32 @@ export const bas = {
...
+        const isInBAS = !_.isEmpty(_.get(process, onBasEnv));
```
bd82, line 42: try using optional chaining here instead. of the `!_.isEmpty(_.get`
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html

src/api.ts
ira-gordin-sap: This is constant.

src/api.ts
```diff
@@ -35,6 +37,32 @@ export const bas = {
...
+        if (!isInBAS) {
+            getLogger().trace("Running in VS Code, so returning null.");
+            return null;
+        }
+        getLogger().trace("Running on BAS.");
+
+        let sap;
+        try {
+            sap = require('@sap/plugin');
```
bd82, line 51: try to use `https://www.npmjs.com/package/optional-require`

src/api.ts
```diff
@@ -35,6 +37,32 @@ export const bas = {
...
+        } catch (error) {
+            getLogger().error("Failed to load @sap/plugin, so returning null.", {error});
+            return null;
+        }
+        getLogger().trace("@sap/plugin successfully loaded.");
+
+        const configuration = await sap.window.configuration();
+        getLogger().trace("Configuration successfully received.", {configuration});
+
+        const parameterValue = _.get(configuration, parameterName, null);
```
bd82, line 61: should we return `null` or `undefined` ?
you ay be able to use optional chaining instead of `_.get`

bd82:
Is there documentation for this `getParameter`? who is the consumer of this API?
How Would they know how to use it / in what scenarios it is relevant?

src/api.ts
bd82: yes, but that was before ESM imports
`import {foo, bar, bamba} from "../blah"`

src/api.ts
bd82: @idoprz WDYT null or undefined?

src/api.ts
```diff
@@ -2,7 +2,9 @@ import * as vscode from 'vscode';
 import { performAction } from './actions/client';
 import { ActionsController } from './actions/controller';
 import { ExecuteAction, SnippetAction, CommandAction, FileAction } from './actions/impl';
+import { getLogger } from './logger/logger';
 export * from "./actions/interfaces";
+import _ = require('lodash');
 
 export const bas = {
```
bd82, line 9: 1. is the type of `bas` defined anywhere?
2. evaluate separating the bas structure and the implementation of each method.
    basically what we have now does not scale, if we will have 20 methods/functions this api.ts file will be 1,000+ line long.

ira-gordin-sap:
Yes, it should be documented. Where is a good place to document it?

src/api.ts
ira-gordin-sap: to move my api to separate file? (thought about it) to create a folder for all apis?

src/api.ts
ira-gordin-sap: When I use optional-require, I don't need to test isInBAS, right? Are you ok I'll remove it?

bd82: not sure, lets sync offline

src/api.ts
bd82: folder for all APIs seems like a good idea.
But start with only your new API (**don't** make this PR into a re-factor)

bd82:
Yes, it should be documented. Where is a good place to document it?

There is a `types` folder in the root of this repo, I am not sure what it currently contains.
We may need to handle the documentation question separately outside this PR.

src/api.ts
ira-gordin-sap: This is from @idoprz 
If the user tries to get value of key that does not exist it returns null. If running in VSCode all keys will return null as value for now.

src/api.ts
ira-gordin-sap: Works perfect for undefined "const parameterValue = configuration?.[parameterName];", but not for null. Or you know a better way?

src/api.ts
ira-gordin-sap: I talked to Ido and it doesn't matter for him null or undefined. 
I do love null, because undefined can come from some bug and null is more sure, but can change if you are sure it is better.

src/api.ts
```diff
@@ -35,6 +37,10 @@ export const bas = {
         return ActionsController.getAction(actionId);
     },
 
+    async getParameter(parameterName: string) : Promise<string | undefined> {
```
bd82, line 40: `  getParameter : getParameter`

src/apis/parameters.ts
```diff
@@ -0,0 +1,29 @@
+//import {isEmpty, get} from 'lodash';
+import { getLogger } from '../logger/logger';
+
+export async function getParameter(parameterName: string) : Promise<string | undefined> {
+    // const onBasEnv = "env.WS_BASE_URL";
```
bd82, line 5: delete commnets...

src/apis/parameters.ts
```diff
@@ -0,0 +1,29 @@
...
+    //const isInBAS = get(process, onBasEnv)?.is;
+    // if (!isInBAS) {
+    //     getLogger().trace("Running in VS Code, so returning undefined.");
+    //     return undefined;
+    // }
+    // getLogger().trace("Running on BAS.");
+
+    const optionalRequire = require("optional-require")(require);
+    const noSapPlugin = "noSapPlugin";
+    const sapPlugin = optionalRequire('@sap/plugin') || noSapPlugin;
```
bd82, line 16:
```javascript
  const noSapPlugin = "NO_SAP_PLUGIN_FOUND";
  const sapPlugin = optionalRequire('@sap/plugin') ?? noSapPlugin;
```

tests/api.spec.ts
```diff
@@ -16,6 +16,7 @@ const testVscode = {
 mockVscode(testVscode, "src/api.ts");
 
 import { bas } from "../src/api";
+import * as parameters from "../src/apis/parameters";
```
bd82, line 19: probably don't need this import with better BAS object definition

tests/api.spec.ts
```diff
@@ -114,6 +111,14 @@ describe("api unit test", () => {
         extensionsMock.expects("getExtension").withExactArgs("myExt").returns(undefined);
         await expect(bas.getExtensionAPI("myExt")).to.be.rejectedWith(`Extension myExt is not loaded`);
     });
+
+    it("getParameter", async() => {
```
bd82, line 115: this test will no longer be needed

tests/apis/parameters.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+import { extend } from "lodash";
+import { createSandbox, stub } from "sinon";
+import { getParameter } from '../../src/apis/parameters';
+
+describe("getParameter", () => {
+    let sandbox: any;
+    const WS_BASE_URL = "WS-BASE-VALUE";
+    const parameterName = "param1";
+
+    before(() => {
```
bd82, line 11: is the sandbox required?

tests/apis/parameters.spec.ts
```diff
@@ -0,0 +1,139 @@
+import { expect } from "chai";
+import { extend } from "lodash";
+import { createSandbox, stub } from "sinon";
+import { getParameter } from '../../src/apis/parameters';
+
+describe("getParameter", () => {
```
bd82, line 6: https://mochajs.org/#bdd (refactor to this style)

tests/apis/parameters.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    //     expect(parameterValue).to.be.undefined;
+    // });
+
+    it("no @sap/plugin loaded ---> returns undefined", async () => {
+        stubEnv({ WS_BASE_URL: WS_BASE_URL });
+        const parameterValue = await getParameter(parameterName);
+        expect(parameterValue).to.be.undefined;
+    });
+
+    it("@sap/plugin loaded, configuration is null ---> returns undefined", async () => {
```
bd82, line 31:
```javascript
context("@sap/plugin loaded", () => {
  describe("when configuration is null", () => {
    let requireMock
    
    before(() => {
      requireMock = require('mock-require');
      const sapPlugin = {
        window: {
          configuration: () => null
        }
      };
      requireMock('@sap/plugin', sapPlugin);
    })

    it("returns undefined", async () => {
      const parameterValue = await getParameter(parameterName);
      expect(parameterValue).to.be.undefined;
    });
    
    after(() => {
      requireMock.stop('@sap/plugin');
    })
  })
})
```

tests/apis/parameters.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+        requireMock.stop('@sap/plugin');
+    });
+
+    it("@sap/plugin loaded, configuration is undefined ---> returns undefined", async () => {
+        stubEnv({ WS_BASE_URL: WS_BASE_URL });
+        const requireMock = require('mock-require');
+
+        const sapPlugin = {
+            window: {
+                configuration: () => undefined
```
bd82, line 53: inspect what is possible according to `window.configuration()` API.
and consider which edge cases need to be checked.

tests/apis/parameters.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+            }
+        };            
+        requireMock('@sap/plugin', sapPlugin);
+
+        const parameterValue = await getParameter(parameterName);
+        expect(parameterValue).to.be.equal(expectedParameterValue);
+        requireMock.stop('@sap/plugin');
+    });
+
+    it("@sap/plugin loaded, parameter name exists in the configuration with undefined value ---> returns the parameter value", async () => {
```
bd82, line 114: like before you may not need to check all edge cases of parameters values (null/undefined) and instead can rely on member access syntax `["foo"]`

tests/apis/parameters.spec.ts
ira-gordin-sap: The API:
  export namespace window {
    export function configuration(): Promise<{ [id: string]: string }>;
  }
They has initialization to empty in the code:
const configuration: { [id: string]: string } = {};
So currently it can be only empty and not null and not undefined.
? behaves the same on null and undefined
So I am keeping only null configuration.

tests/apis/parameters.spec.ts
```diff
@@ -0,0 +1,94 @@
+import * as vscode from "vscode";
```
bd82, line 1: i think this is import is unused and will cause runtime errors during tests

tests/apis/parameters.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+            });
+            
+            after(() => {
+                requireMock.stop('@sap/plugin');
+            });
+        });
+    });
+      
+    context("@sap/plugin loaded", () => {
+        // no test for configuration containing other parameters, because [] behaves the same
```
bd82, line 42: [] --> to 'member access'

because `[]` can also reference an array or destructuring in JavaScript syntax.

bd82 (approved):

.mocharc.yaml
```diff
@@ -0,0 +1,6 @@
+require:
```
bd82, line 1: use a `.js` file instead of yaml for .mocharc
as it would allow better re-use / combining

package.json
```diff
@@ -56,14 +56,15 @@
 		"compile": "tsc -p ./",
 		"watch": "tsc -watch -p ./",
 		"pretest": "npm run compile",
-		"test": "nyc mocha -p tsconfig.json --recursive --no-parallel tests/**/*.spec.ts",
+		"test": "nyc mocha -p tsconfig.json --config ./.mocharc.yaml --no-parallel",
```
bd82, line 59: I think the mocharc file should be automatically loaded if it is in the same folder

bd82 (approved):
you can fix the mocha configuration in a different PR

slavik-lvovsky (approved):

.mocharc.yaml
ira-gordin-sap: this is .js now

PR #47, opened 2021-04-21
chore: fix tests mocha configuration

ira-gordin-sap:


.mocharc.js
```diff
@@ -3,5 +3,5 @@ const chaiAsPromised = require("chai-as-promised");
 
 chai.use(chaiAsPromised);
 module.exports = {
-    
+  spec: "./tests/**/*.spec.ts",
```
bd82, line 6: the `ts-node` is no longer required?

ira-gordin-sap: We didn't have `ts-node`.
.mocharc.yaml was some try to fix the issue.
Before it, we had .mocharc.js, but we missed this spec configuration.

bd82 (approved):

NetaMolcho (approved):

PR #48, opened 2021-04-26
feat: perform actions from url in activate extension

NetaMolcho:


src/actions/controller.ts
```diff
@@ -23,17 +27,49 @@ export class ActionsController {
...
+      const actionsList = [];
+
+      const actionParam = await getParameter("action");
+      const actionsParam = await getParameter("actions");
+      logger.trace(`configuration - action= ${actionParam}, actions= ${actionsParam}`);
+      const actionIds = actionParam?.split(",") || [];
+      let actionsIds = actionsParam?.split(",") || [];
+      actionsIds = actionsIds.concat(actionIds);
+      actionsIds = uniq(actionsIds);
+      if (actionsIds.length > 0) {
```
slavik-lvovsky, line 41: this if does not help, you can remove it

src/actions/controller.ts
```diff
@@ -23,17 +27,49 @@ export class ActionsController {
...
+          } else {
+            logger.trace(`action ${actionId} not found`);
           }
-          actionsSettings.update("actions", []);
         }
+        ActionsController.performActions(actionsList);
+      }
+    }
+
+    public static performActions(actionsList: any[]) {
```
slavik-lvovsky, line 55: it should be private

src/actions/controller.ts
```diff
@@ -23,17 +27,49 @@ export class ActionsController {
...
+          const action = ActionsController.getAction(actionId);
+          if (action){
+            logger.trace(`action ${actionId} found`, {action});
+            actionsList.push(action);
+          } else {
+            logger.trace(`action ${actionId} not found`);
           }
-          actionsSettings.update("actions", []);
         }
+        ActionsController.performActions(actionsList);
```
slavik-lvovsky, line 51: just call _performAction for each action, no need to create another method

slavik-lvovsky (changes requested):

src/actions/controller.ts
NetaMolcho: done

src/actions/controller.ts
NetaMolcho: done

src/actions/controller.ts
NetaMolcho: done

src/actions/controller.ts
```diff
@@ -23,17 +27,44 @@ export class ActionsController {
       }
     }
 
+    public static async performActionsFromParams() {
+      const logger = getLogger().getChildLogger({label: ActionsController.loggerLabel});
```
slavik-lvovsky, line 31: this logger is used twice in this file, make logger member instead of loggerLabel

src/actions/controller.ts
```diff
@@ -23,17 +27,44 @@ export class ActionsController {
...
-              `performing action ${action.name} of type ${action.constructor.name}`
-            );
-            _performAction(action);
-          }
-          actionsSettings.update("actions", []);
+      const logger = getLogger().getChildLogger({label: ActionsController.loggerLabel});
+      const actionsSettings = vscode.workspace.getConfiguration();
+      const actionsList: any[] | undefined = actionsSettings.get("actions");
+      if (actionsList && actionsList.length) {
+        for (const action of actionsList) {
```
slavik-lvovsky, line 59: by using _.forEach instead of for you will not need "if"

tests/extension.spec.ts
```diff
@@ -104,4 +103,87 @@ describe("extension unit test", () => {
         extension.deactivate();
     });
 
+    context("activate with actionId as parameters in the URL", () => {
```
slavik-lvovsky, line 106: Can you create needed tests in controller.spec.ts ? Unit tests should be in controller.spec.ts. Here it looks like integration test

slavik-lvovsky (changes requested):

src/actions/controller.ts
```diff
@@ -23,17 +27,49 @@ export class ActionsController {
...
-          for (const action of actionsList) {
-            console.log(
-              `performing action ${action.name} of type ${action.constructor.name}`
-            );
-            _performAction(action);
+    public static async performActionsFromParams() {
+      const logger = getLogger().getChildLogger({label: "performActionsFromParams"});
+      const actionsList = [];
+
+      const actionParam = await getParameter("action");
```
slavik-lvovsky, line 33: need to take only first action id from "action" parameter

slavik-lvovsky (approved):

PR #49, opened 2021-04-28
feat: project type poc

bd82:
## MVP
- [ ] Project Type library "mock"
- [ ] FIll VSCode `maps` according to project type mock values.
- [ ] Extension with context menu button using `in` condition. 

## Bonus Scenarios
- [ ] passing library plugin via VSCode extensions API.
- [ ] ???


src/api.ts
```diff
@@ -12,7 +11,7 @@ import {
 import { getParameter } from "./apis/parameters";
 import { getLogger } from "./logger/logger";
 
-const basToolkitAPI: Omit<BasToolkit, "workspaceAPI"> = {
+const basToolkitAPI = {
```
bd82, line 14: Why did you remove the type signature?
We are losing type checking on the object literal here...

src/project-type/context-state.ts
```diff
@@ -50,9 +51,7 @@ export function refreshAllVSCodeContext(): void {
   });
 }
 
-export async function initTagsContexts(
-  workspaceAPI: WorkspaceApi
-): Promise<void> {
+export async function initTagsContexts(workspaceAPI: BasWorkspaceApi): Promise<void> {
```
bd82, line 54: why does our **internal** code needs to use the **read-only** interface?

src/project-type/workspace-instance.ts
```diff
@@ -1,18 +1,38 @@
...
 
+let basWSAPI: BasWorkspaceApi;
 /**
  * @param WorkspaceImpl - constructor for the `WorkspaceApi`
  *                        dependency injection is used to enable easier testing
  */
-export function initWorkspaceAPI(WorkspaceImpl: { new (): WorkspaceApi }) {
-  workspaceAPIProxy = new WorkspaceImpl();
-  // TODO: implement "READ-ONLY" proxy
+export function getBasWorkspaceAPI(WorkspaceImpl: { new(): WorkspaceApi }): BasWorkspaceApi {
```
bd82, line 10: The split of the `init` and the `getter` sounds logical.
But because of the DI for the constructor it means everyone who calls `getBasWorkspaceAPI`
now needs to know how to provide the constructor so the end result may be worse.

bd82: It also prevents us from ever changing the `basWSAPI` instance (e.g in testing)
As with this code it would only be initialized **once**.

src/project-type/workspace-instance.ts
```diff
@@ -1,18 +1,38 @@
...
-export function getWorkspaceAPI(): WorkspaceApi {
-  return workspaceAPIProxy;
+
+function initWorkspaceImpl(workspaceImpl: WorkspaceApi): void {
+  workspaceImpl.startWatch();
 }
+
+function createWorkspaceReadOnlyProxy(workspaceImpl: WorkspaceApi): BasWorkspaceApi {
+  const basWSAPI: BasWorkspaceApi = {
+    getProjects: (tag?: Tag): Promise<ProjectApi[]> => workspaceImpl.getProjects(tag),
```
bd82, line 29: I wonder if there is a way for us to avoid copying the signatures of the `WorkspaceApi`

tests/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { mockVscode } from "../mockUtil";
+
+const testVscode = {
+    workspace: {
+        workspaceFolders: () => []
+    }
+};
+
+mockVscode(testVscode, "src/project-type/workspace-instance.ts");
```
bd82, line 13: Can we avoid mocking VSCode here by using DI ? (Pure unit tests)

tests/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+import { getBasWorkspaceAPI } from "../../src/project-type/workspace-instance";
+
+describe.skip("getBasWorkspaceAPI", () => {
+    it("multiple bas workspace instances should be equal", () => {
```
bd82, line 17: "equal" --> "identical"

or "getBasWorkspaceAPI will always return the **same** instance"

tests/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+        const basWsInstances = [];
+        for (let i = 0; i < 5; i++) {
+            basWsInstances.push(getBasWorkspaceAPI(WorkspaceImpl));
```
bd82, line 20: This is completely **subjective**, but I would do a one liner to initialize this `basWsInstances` array.
```javascript
const {range, map} = require("lodash")
const basWsInstances = map(range(0,5), () => getBasWorkspaceAPI(WorkspaceImpl))
```

If we remove the constructor parameter from `getBasWorkspaceAPI` it becomes even shorter.

```javascript
const {range, map} = require("lodash")
const basWsInstances = map(range(0,5), getBasWorkspaceAPI)
```

The reason it is **subjective** is that is general style differences between imperative and functional programmatic styles.
On the other hand we do use lodash semi functional style a-lot so it may fit our project.

tests/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+        }
+        expect(uniq(basWsInstances)).to.have.lengthOf(1);
```
bd82, line 22: It is really nice you are testing for multiple (5) invocation results.
However this does make the assertion less understandable (how vs what).
It should be possible to write something like:
```javascript
const firstBasWsApi = getBasWorkspaceAPI();
expect(basWsInstances).to.each.equal(firstBasWsApi);
```

but that would require adding another chai plugin which is likely overkill if we only do this assertion once:
- https://www.chaijs.com/plugins/chai-each/

types/api.d.ts
```diff
@@ -1,6 +1,8 @@
 import { Uri } from "vscode";
 import { WorkspaceApi } from "@sap/project-api";
 
+declare type BasWorkspaceApi = Pick<WorkspaceApi, "getProjects" | "getProjectUris" | "onWorkspaceChanged">;
```
bd82, line 4: using Pick is better than Omit as we explicitly control what our API includes.
👍

tests/project-type/workspace-instance.spec.ts
slavik-lvovsky: I will add the plugin once we have one more similar test

src/api.ts
slavik-lvovsky: Done

src/project-type/context-state.ts
slavik-lvovsky: Done

src/project-type/workspace-instance.ts
slavik-lvovsky: Done

src/project-type/workspace-instance.ts
slavik-lvovsky: Just to disable eslint error, please check but I can revert to the previous logic

tests/project-type/workspace-instance.spec.ts
slavik-lvovsky: I did not find. I added mock objects to other tests

tests/project-type/workspace-instance.spec.ts
slavik-lvovsky: Done

tests/project-type/workspace-instance.spec.ts
slavik-lvovsky: Done

src/project-type/workspace-instance.ts
```diff
@@ -26,12 +22,12 @@ function initWorkspaceImpl(workspaceImpl: WorkspaceApi): void {
 
 function createWorkspaceReadOnlyProxy(workspaceImpl: WorkspaceApi): BasWorkspaceApi {
   const basWSAPI: BasWorkspaceApi = {
-    getProjects: (tag?: Tag): Promise<ProjectApi[]> => workspaceImpl.getProjects(tag),
-    
-    getProjectUris: (): Promise<Uri[]> => workspaceImpl.getProjectUris(),
-
-    onWorkspaceChanged: (handler: (event: string, folders: WorkspaceFolder[]) => void): void => 
-      workspaceImpl.onWorkspaceChanged(handler)
+    // eslint-disable-next-line @typescript-eslint/unbound-method
```
bd82, line 25: You can disable a whole block (start/stop) in ESLInt.
That would be more useful particularly if more methods would be added to the BasWorkspaceAPi

tests/project-type/workspace-instance.spec.ts
```diff
@@ -1,24 +1,20 @@
-import { WorkspaceImpl } from "@sap/project-api";
 import { expect } from "chai";
-import { uniq } from "lodash";
-
+import * as _ from "lodash";
```
bd82, line 2: **subjective**

while this does not matter as much for test code it is better to import only the names you need
`import { map, range }  from "lodash";`

Reasons:
- bundling tools can better do tree- shaking 
   - https://webpack.js.org/guides/tree-shaking/
- the identifier `_` (underscore) is useful as an ID for very short lambda functions

tests/project-type/workspace-instance.spec.ts
```diff
@@ -1,24 +1,20 @@
...
 
-describe.skip("getBasWorkspaceAPI", () => {
-    it("multiple bas workspace instances should be equal", () => {
-        const basWsInstances = [];
-        for (let i = 0; i < 5; i++) {
-            basWsInstances.push(getBasWorkspaceAPI(WorkspaceImpl));
-        }
-        expect(uniq(basWsInstances)).to.have.lengthOf(1);
+describe("getBasWorkspaceAPI", () => {
+    it("multiple bas workspace instances should be identical", () => {
```
bd82, line 16: I am thinking about this again and I am not sure:
- Do we really need this test?
- Does it matter if the `getBasWorkspaceAPI` returns the same instance?
- What is important is that for all our consumers via `vscode.extensions.getExtension("bas-toolkit").exports.workspaceApi` gets the same instance.

Do you remember why we did this test? I think I suggested it for some reason but it does not make sense for me now.

src/extension.ts
```diff
@@ -28,9 +28,9 @@ export function activate(context: ExtensionContext) {
   const logger = getLogger().getChildLogger({ label: "activate" });
   logger.info("The App-Studio-Toolkit Extension is active.");
 
-  const workspaceAPI: BasWorkspaceApi = getBasWorkspaceAPI(WorkspaceImpl);
-  void initTagsContexts(workspaceAPI);
-  void initProjectTypeWatchers(workspaceAPI);
+  const workspaceAPI: BasWorkspaceApi = getBasWorkspaceAPI();
+  void initTagsContexts(workspaceAPI as WorkspaceApi);
+  void initProjectTypeWatchers(workspaceAPI as WorkspaceApi);
```
bd82, line 33: There is a bit of confusion here between the internal (implementation) and external (api) concerns.
- Internally we may need the full `WorkspaceApi` instance and all its methods.
- Externally we are limiting the available APIs.

These two `initXYZ` functions are internal, so why are they working with the **External** proxy?
Also note we don't really need to save a reference to the proxied instance.

tests/project-type/workspace-instance.spec.ts
bd82: needing new mocks for existing mocks based tests makes sense.
but why does this test needs mocks?

src/project-type/workspace-instance.ts
```diff
@@ -0,0 +1,34 @@
+import { WorkspaceApi, WorkspaceImpl } from "@sap/project-api";
```
bd82, line 1: did not we want to avoid importing `WorkspaceImpl` class here?

src/project-type/workspace-instance.ts
```diff
@@ -0,0 +1,34 @@
...
+import { BasWorkspaceApi } from "../../types/api";
+
+let basWSAPI: BasWorkspaceApi;
```
bd82, line 4: - I think we need to keep a ref to the `WorkspaceImpl` for our internal logic
- I am not sure we need to keep a ref to the "read-only" version.

src/project-type/workspace-instance.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function getBasWorkspaceAPI(): BasWorkspaceApi {
+  if (!basWSAPI) {
+    const workspaceImpl = new WorkspaceImpl();
```
bd82, line 8: I think we will need to separate the concerns in this module so it would be easily testable.

lets discuss offline

src/project-type/workspace-instance.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    basWSAPI = createWorkspaceReadOnlyProxy(workspaceImpl);
+  }
+
+  return basWSAPI;
+}
+
+
+function initWorkspaceImpl(workspaceImpl: WorkspaceApi): void {
+  workspaceImpl.startWatch();
```
bd82, line 20: Should `startWatch()`  happen after `getBasWorkspaceAPI()` is called?
But we need the "watching" to work also for internal logic (e.g the VSCode contexts watching logic).
So maybe it should be de-coupled from the **External** BasWsApi" logic

tests/project-type/workspace-instance.spec.ts
slavik-lvovsky: I think that vscode already helps us to get same instance of basTollkitAPI.
I have added date property to basTollkitAPI in the createBasToolkitAPI method (Date.now())
In the sample example I have made 2 separate calls to app-studio-toolkit extension.
The dates property were equal.

PR #57, opened 2021-05-09
chore: improve apis and examples

slavik-lvovsky:


bd82:
if APIs where changed its not a chore, it may even be a breaking change...

src/actions/actionsConfig.ts
```diff
@@ -0,0 +1,11 @@
+import { workspace } from 'vscode';
+
+const key = "actions";
+
+export const get = (): string[] => {
```
bd82, line 5: 1. why do we need a const to expose a function, can't we just export a function directly?

src/actions/actionsFactory.ts
```diff
@@ -1,40 +1,89 @@
 import { Uri } from "vscode";
-import { IAction, ICommandAction, ActionType, IFileAction, ActionJsonKey } from './interfaces';
-import { CommandAction, FileAction } from './impl';
+import { IAction, ICommandAction, ActionType, IFileAction, ActionJsonKey, ISnippetAction } from './interfaces';
+import { CommandAction, FileAction, SnippetAction } from './impl';
+import _ = require("lodash");
```
bd82, line 4: I recommend you only import what you actually need from lodash.

src/actions/actionsFactory.ts
```diff
@@ -1,40 +1,89 @@
...
+
+const getNameProp = (fromSettings: boolean): string => {
```
bd82, line 6: why `const` to expose a function?

src/actions/actionsFactory.ts
```diff
@@ -1,40 +1,89 @@
...
+    return fromSettings ? "name" : ActionJsonKey.CommandName;
+};
+const getParamsProp = (fromSettings: boolean): string => {
```
bd82, line 9: same as above

src/actions/actionsFactory.ts
```diff
@@ -1,40 +1,89 @@
...
-                        commandAction.name = commandName;
-                    } else {
-                        throw new Error(`${ActionJsonKey.CommandName} is missing for actionType=${actionType}`);
-                    }
-                    const commandParams = jsonAction[ActionJsonKey.CommandParams];
-                    if (commandParams) {
-                        commandAction.params = commandParams;
-                    }
-                    return commandAction;
+                const commandAction: ICommandAction = new CommandAction();
```
bd82, line 21: please Increase readability be extracting each handler to a separate sub function

```javascript
switch (actionType) {
            case ActionType.Command: {
                return handleCommandAction(args...)
            }
            case ActionType.Snippet: {
                return handleSnippetAcion(args...)
            }
            case ActionType.File: {
                return handleFileAction(args...)
            }
            default:
                throw new Error(`Action with ${ActionJsonKey.ActionType}=${actionType} could not be created from json file`);
        }
```

src/actions/client.ts
```diff
@@ -1,23 +1,24 @@
...
-import { _performAction } from "./performer";
-import { getLogger } from "../logger/logger";
+import { workspace, ConfigurationTarget } from 'vscode';
+import { IAction } from './interfaces';
+import { _performAction } from './performer';
+import { getLogger } from '../logger/logger';
 
 
-export function performAction(action: IAction, options?: any): Thenable<void> {
+export const performAction = (action: IAction, options?: any): Thenable<void> => {
```
bd82, line 7: why are functions wrapped in a const?

src/actions/client.ts
```diff
@@ -1,23 +1,24 @@
...
-}
+};
 
-const logger = getLogger().getChildLogger({label: "client"});
+const logger = getLogger().getChildLogger({ label: "client" });
 
 /** Schedule an action for execution after restart */
-function _scheduleAction(action: IAction): Thenable<void> {
-    const actionsSettings = vscode.workspace.getConfiguration();
+const _scheduleAction = async (action: IAction): Promise<void> => {
```
bd82, line 14: why are functions wrapped in a const?

src/actions/controller.ts
```diff
@@ -1,47 +1,61 @@
-import * as vscode from "vscode";
+import { extensions } from "vscode";
 import { getLogger } from "../logger/logger";
 import { IAction } from "./interfaces";
 import { _performAction } from "./performer";
 import { getParameter } from '../apis/parameters';
-import { forEach, uniq, get, split, compact } from "lodash";
+import { ActionsFactory } from './actionsFactory';
+import * as _ from "lodash";
```
bd82, line 7: please use specific named imports from lodash

src/actions/controller.ts
```diff
@@ -1,47 +1,61 @@
...
 
-  public static loadActions() {
-    vscode.extensions.all.forEach(extension => {
-      const extActions = get(extension, "packageJSON.BASContributes.actions", []);
-      extActions.forEach((action: IAction) => {
-        ActionsController.actions.push(action);
+  public static loadContributedActions() {
+    const logger = getLogger().getChildLogger({ label: "loadContributedActions" });
+    extensions.all.forEach(extension => {
+      const extensionActions = _.get(extension, "packageJSON.BASContributes.actions", []);
```
bd82, line 17: you can probably replace the `_.get(extension, "packageJSON.BASContributes.actions", []);` pattern
with modern javascript
```javascript
extension?.packageJSON?.BASContributes?.actions ?? []
```

src/actions/controller.ts
```diff
@@ -1,47 +1,61 @@
...
   private static readonly actions: IAction[] = [];
 
-  public static loadActions() {
-    vscode.extensions.all.forEach(extension => {
-      const extActions = get(extension, "packageJSON.BASContributes.actions", []);
-      extActions.forEach((action: IAction) => {
-        ActionsController.actions.push(action);
+  public static loadContributedActions() {
+    const logger = getLogger().getChildLogger({ label: "loadContributedActions" });
+    extensions.all.forEach(extension => {
```
bd82, line 16: avoid using built in array functions in a project which uses lodash.
The reasoning is that it creates **two ways** to read the high order functions.

```javascript
[].forEach(() => {})
forEach([], () => {}
```

src/actions/controller.ts
bd82: This also allows using the `_` identifier a short name in simple functions.

```javascript
forEach(ids,  _ => console.log(_))
```

src/actions/controller.ts
```diff
@@ -1,47 +1,61 @@
...
       }
     });
   }
 
   public static performScheduledActions() {
-    const actionsSettings = vscode.workspace.getConfiguration();
-    const actionsList: any[] = actionsSettings.get("actions", []);
-    forEach(actionsList, action => {
-      void _performAction(action);
+    const logger = getLogger().getChildLogger({ label: "performScheduledActions" });
```
bd82, line 49: why do we need a childLogger for this one function?
Could we not just add an argument on the logging invocation?
The logger should be able to accept an **object** as the arguments.

src/actions/performer.ts
```diff
@@ -1,34 +1,28 @@
...
-  ISnippetAction } from "./interfaces";
+import { commands, ViewColumn } from 'vscode';
+import { getLogger } from '../logger/logger';
+import {
+  ActionJsonKey, ActionType,
+  IAction, ICommandAction,
+  IExecuteAction, IFileAction,
+  ISnippetAction
+} from './interfaces';
+import { get } from 'lodash';
```
bd82, line 9: order import:
-  3rd party libs before relative imports.

src/actions/performer.ts
```diff
@@ -1,34 +1,28 @@
...
 
 export async function _performAction(action: IAction): Promise<any> {
-  const logger = getLogger().getChildLogger({label: "_performAction"});
+  const logger = getLogger().getChildLogger({ label: "_performAction" });
```
bd82, line 12: using childLogger for custom method name logging may be overkill

tests/actionsFactory.spec.ts
```diff
@@ -58,6 +58,51 @@ describe("actionsFactory test", () => {
...
+                id: "id",
+                snippetName: "name",
+                contributorId: "contributorId",
+                context: {},
+                isNonInteractive: true
+            };
+            const action = ActionsFactory.createAction(actionJson);
+            expect(action instanceof SnippetAction).to.be.true;
+            expect((action as SnippetAction).snippetName).to.be.equal("name");
+            expect((action as SnippetAction).contributorId).to.be.deep.equal("contributorId");
```
bd82, line 74: `to.be.deep.equal` is not proper english
I think it should be `to.deep.equal`

note it does not change the runtime, it is style/grammar only

tests/actionsFactory.spec.ts
```diff
@@ -58,6 +58,51 @@ describe("actionsFactory test", () => {
...
+        });
+
+        it("fails without snippetName", () => {
+            const actionJson = {
+                [ActionJsonKey.ActionType]: ActionType.Snippet,
+                contributorId: "contributorId"
+            };
+            expect(() => ActionsFactory.createAction(actionJson)).to.throw(`snippetName is missing for actionType=${ActionType.Snippet}`);
```
bd82, line 94: You can pass a regexp to `to.throw` so you won't assert on the exact string but instead on some pattern
to make it easier to maintain if/when the error message changes.

```javascript
var badFn = function () { throw new TypeError('Illegal salmon!'); };
expect(badFn).to.throw(/salmon/);
```

tests/controller.spec.ts
```diff
@@ -0,0 +1,169 @@
+import { expect } from "chai";
+import { mockVscode } from "./mockUtil";
+import { SinonSandbox, SinonMock, createSandbox } from "sinon";
+import * as _ from "lodash";
```
bd82, line 4: use specific named imports from lodash

tsconfig.json
```diff
@@ -12,5 +12,5 @@
 		"rootDirs" : ["./src", "./tests"],
 		"strict": true
 	},
-	"include": ["src/**/*.ts", "tests/**/*.ts", "api.d.ts"]
+	"include": ["src/**/*.ts", "tests/**/*.ts", "api.d.ts", "src/utils"]
```
bd82, line 15: why is `"src/utils"` needed here? does not `"src/**/*.ts"` include it already?

bd82:
In the future we need to try and separate PRs like this to several different PRs.
It contains multiple topics
- upgrading examples..
- code refactoring
- linting improvements.
- build/bug fixes in types package
- ...

Also it may not a *chore*. although we do not use conventional commit standard in this repo yet.

src/actions/actionsConfig.ts
bd82: discuss later

src/actions/actionsFactory.ts
bd82: discuss later

src/actions/actionsFactory.ts
bd82: discuss later

src/actions/actionsFactory.ts
bd82: please resolve

bd82 (approved):
please resolve things we discussed online before merge

PR #71, opened 2021-05-24
feat: passing parameters to actions via the URL

NetaMolcho:


src/actions/controller.ts
```diff
@@ -29,16 +29,38 @@ export class ActionsController {
     return ActionsController.actions.find(action => action.id === id);
   }
 
-  public static async performActionsFromParams() {
+  public static async performActionsFromURL() {
     const actionsParam = await getParameter("actions");
-    const actionsIds = uniq(compact(split(actionsParam, ",")));
-    getLogger().trace(`configuration - actionsIds= ${actionsIds}`, { method: "performActionsFromParams" });
+    getLogger().trace(`actionsParam= ${actionsParam}`, { method: "performActionsFromURL" });
+    if (actionsParam?.startsWith("%5B") || actionsParam?.startsWith("[")) {
```
bd82, line 35: can you add a one line comment how the action parameter looks in each branch, e.g:
`// actions=[{},{}]`
`// actions=foo,bar`

src/actions/controller.ts
```diff
@@ -29,16 +29,38 @@ export class ActionsController {
...
+      ActionsController.perfomFullActions(actionsParam);
```
bd82, line 36: I will try to think of a new name to emphasize this action is:
- defined on the fly.
- fully specified.
-  does not match an existing action

bd82: maybe in-lined?

tests/controller.spec.ts
```diff
@@ -166,4 +166,44 @@ describe("controller unit test", () => {
             }
         });
     });
+
+    describe('perfomFullActions', () => {
+        let requireMock: any;
+        before(() => {
```
bd82, line 172: should we use `before**each**` and `after**each**` to increase test case isolation?

src/actions/controller.ts
NetaMolcho: renamed to perfomInlinedActions

tests/controller.spec.ts
NetaMolcho: This only matters for the first unit test

src/actions/controller.ts
```diff
@@ -29,16 +28,63 @@ export class ActionsController {
...
-    getLogger().trace(`configuration - actionsIds= ${actionsIds}`, { method: "performActionsFromParams" });
+    if (actionsParam === undefined) {
+      return;
+    }
+    const decodedActionsParam = decodeURI(actionsParam);
+    getLogger().trace(`decodedActionsParam= ${decodedActionsParam}`, { method: "performActionsFromURL" });
+    const mode = ActionsController.detectActionMode(decodedActionsParam);
+    switch (mode) {
+      case "IDs": {
+        const actionsIds = uniq(compact(split(decodedActionsParam, ",")));
```
bd82, line 41: we need to trim the array of IDs

src/actions/controller.ts
```diff
@@ -29,16 +28,63 @@ export class ActionsController {
...
+        ActionsController.performActionsIds(actionsIds);
+        break;
+      }
+      case "Inlined": {
+        ActionsController.perfomInlinedActions(decodedActionsParam.trim());
```
bd82, line 46: - we don't need to trim here, JSON.parse() does it.
- we should pass the decoded string  here

src/actions/controller.ts
bd82: I saw you did it inside: `performActionsIds`
but then we have the "processing" of input split into two places.
I'd rather do it in one place only

src/actions/controller.ts
```diff
@@ -29,16 +28,63 @@ export class ActionsController {
...
+      //actions=openSettings,openGoogle
+    }
+    return "IDs";
+  }
+
+  private static performActionsIds(actionsIds: string[]) {
+    getLogger().trace(`actionsIds= ${actionsIds}`, { method: "performActionsIds" });
       forEach(actionsIds, async actionId => {
-      const action = ActionsController.getAction(actionId);
+      const action = ActionsController.getAction(actionId.trim());
```
bd82, line 70: do the trimming with the other processing (uniq/concat/...) so the "processing" is handled in a **single** place.

src/actions/controller.ts
```diff
@@ -29,16 +28,63 @@ export class ActionsController {
...
+      }
+    }
+    catch (e) {
+      // actionsIDsMode
+      //actions=openSettings,openGoogle
+    }
+    return "IDs";
+  }
+
+  private static performActionsIds(actionsIds: string[]) {
```
bd82, line 67: what do you think about naming this function: performActions**By**Ids

So we have:
- perform**Inlined**Actions
- performActions**ByIds**

Which makes more sense from English grammar perspective.

src/actions/controller.ts
bd82: subjective because we immediately use it.

tests/controller.spec.ts
bd82: We still need good isolation between tests to avoid none deterministic behavior.
For example if we do `.only` for the second test these `before` and `after` blocks would still run
can we guarantee they would not affect the results?

src/actions/controller.ts
bd82: decided this is subjective 👍

bd82 (approved):

PR #90, opened 2021-07-12
chore: add support for islcapenables api

idantr:


src/api.ts
```diff
@@ -32,15 +33,16 @@ export const bas:BasToolkit = {
         });
     },
 
-    getAction: (actionId: string) => ActionsController.getAction(actionId),
-    getParameter,
-    performAction,
+    //getAction: (actionId: string) => ActionsController.getAction(actionId),
+    //getParameter,
+    //performAction,
+    isLCAPEnabled,
```
NetaMolcho, line 40: Why do you mark these functions as a comment? they are used in many places, including other repositories

src/api.ts
```diff
@@ -32,15 +33,16 @@ export const bas:BasToolkit = {
...
-    actions: {
+    /*actions: {
         performAction,
         ExecuteAction,
         SnippetAction,
         CommandAction,
         FileAction
-    }
+    }*/
 };
```
NetaMolcho, line 48: same here

src/api.ts
idantr: as far as i understand from rotem i need to remove all the action and getparameters calls
@idoprz can you also got over the code

NetaMolcho (approved):

bd82:
There are multiple problems with this with this PR:

- Removing an API ('getParameter') is a **Breaking Change** (Major Version) not a minor version change.
  - https://github.com/SAP/app-studio-toolkit/pull/90/files#diff-99210edf83ca3c5dffc806ab8d5329433ef0faf703ec68d4cc59f000f6e82091R3
- The usage of 'getParameter' was not removed from the example, thus it currently fails compilation.
  - https://github.com/SAP/app-studio-toolkit/blob/6027b6e65a0f953bd9d1d2de91e1e94c5086f3cb/sample-action-client/src/extension.ts#L73
- Commenting out a line of code is **not a valid manner** in which to remove a function from a public API.
  - https://github.com/SAP/app-studio-toolkit/blob/6027b6e65a0f953bd9d1d2de91e1e94c5086f3cb/types/api.d.ts#L28-L37
- This PR is a `feat` not a `chore` and the commit message includes redundant dev process info
  - See:
image

I am in the middle of a large unrelated refactor in this repo to a mono-repo, please do not push any new commits
to resolve these issues, I will handle it after the refactor.

PR #126, opened 2021-10-13
fix: execute actions defined in workspace folders and workspace settings

slavik-lvovsky:


packages/app-studio-toolkit/src/actions/actionsConfig.ts
```diff
@@ -1,11 +1,31 @@
...
-export const get = (): string[] => {
-  return workspace.getConfiguration().get(key, []);
+const addActions = (actions: any[], config: WorkspaceConfiguration): void => {
+  const configActions = config.get<any[]>(key, []);
+  actions.splice(actions.length, 0, ...configActions);
 };
 
-export const clear = (): Thenable<void> => {
-  return workspace.getConfiguration().update(key, []);
+export const get = (): any[] => {
```
bd82, line 11: overuse of `any`

packages/app-studio-toolkit/src/actions/actionsConfig.ts
```diff
@@ -1,11 +1,31 @@
...
+  const actions: any[] = [];
+
+  workspace.workspaceFolders?.forEach((wsFolder) => {
```
bd82, line 14: it is better to use lodash's `forEach` because the built-in high order functions in JS are very limited and insufficient and if we use both them and lodash's style we will inconsistent code style. 

So might as well use the style of the larger/more complete set of utilities (lodash)

packages/app-studio-toolkit/src/actions/actionsConfig.ts
```diff
@@ -1,11 +1,31 @@
...
+    addActions(actions, workspace.getConfiguration(undefined, wsFolder.uri));
+  });
+
+  addActions(actions, workspace.getConfiguration());
+
+  return uniqWith(actions, isEqual);
+};
+
+export const clear = (): void => {
+  workspace.workspaceFolders?.forEach((wsFolder) => {
```
bd82, line 24: see comment above on lodash's forEach

PR #150, opened 2021-11-14
feat: npm dependencies validations

slavik-lvovsky:


bd82:
lets target this to a new feature branch (not main) so we can incrementally merge sub-features.

packages/npm-dependencies-validation/.mocharc.js
```diff
@@ -0,0 +1,10 @@
+const chai = require("chai");
+const chaiAsPromised = require("chai-as-promised");
+
+chai.use(chaiAsPromised);
+
+module.exports = {
+  require: ["source-map-support/register"],
+  spec: "./dist/test/**/*spec.js",
+  timeout: 5000,
```
bd82, line 9: is 5000ms timeout really needed?
are the tests slow because of npm operations?
if so even higher timeout may be needed in container/vms (github actions).

Also are there only a few tests that are very slow? or are all tests slow?
Because a per test timeout can be specified in mocha.

packages/npm-dependencies-validation/.npmignore
```diff
@@ -0,0 +1,6 @@
+node_modules
```
bd82, line 1: if you include the sources and the source maps
it will allow easier debugging for consumers.

bd82: Also I am not sure `.npmignore` is needed
could we instead just specify the files we want to package **explicitly** in the package.json "files" section?

packages/npm-dependencies-validation/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@sap-devx/npm-dependencies-validation",
+  "version": "1.3.0",
+  "private": true,
+  "description": "",
```
bd82, line 5: short description needed

packages/npm-dependencies-validation/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "author": "SAP SE",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/SAP/app-studio-toolkit"
```
bd82, line 10: is there a way to link to a sub-folder in a repo

packages/npm-dependencies-validation/package.json
```diff
@@ -0,0 +1,20 @@
...
+  },
+  "main": "./dist/src/index",
+  "types": "./dist/src/index.d.ts",
```
bd82, line 13: we should consider if we want separate types declarations or re-use the generated types signatures (I am not sure...)

packages/npm-dependencies-validation/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "scripts": {
+    "test": "mocha",
+    "compile:watch": "tsc -p ./ -watch",
```
bd82, line 16: the `watch` script may be redundant as we use `compile:watch` from the root of the repo
which compiles **all** the projects.

packages/npm-dependencies-validation/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "author": "SAP SE",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/SAP/app-studio-toolkit"
+  },
+  "main": "./dist/src/index",
+  "types": "./dist/src/index.d.ts",
+  "scripts": {
+    "test": "mocha",
```
bd82, line 15: there should be a `ci` script which does the full build for this sub-package.
This is the script that will get called from the root of the monorepo
without it this sub-package will not get built

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,90 @@
+import { spawnCommand } from "./npmUtil";
+import { resolve, join } from "path";
+import { access, readFile } from "fs/promises";
+import { constants } from "fs";
+import { DependencyIssue, IssueType } from "./types";
+
+const LS_DEPS = ["ls", "--depth=0"]; // dependencies and extraneous modules
```
bd82, line 7: should theses two constants be with an "ARGS" suffix instead of "DEV"?

packages/vscode-dependencies-validation/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@sap-devx/vscode-dependencies-validation",
```
bd82, line 2: i do not believe vscode extension names can include scopes (@sap-devx)

packages/vscode-dependencies-validation/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "main": "./dist/extension.js",
+  "contributes": {
+    "commands": [
+      {
+        "command": "vscode-dependencies-validation.displayIssues",
+        "title": "Display Issues"
+      }
+    ]
+  },
+  "scripts": {
```
bd82, line 25: instead of generating from a template, I recommend you copy and modify 
from the existing VSCode ext in this mono-repo.

That way we will have a consistent project structure/scripts/flows

packages/vscode-dependencies-validation/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "glob": "^7.1.7",
+    "mocha": "^9.1.3",
+    "typescript": "^4.4.4",
+    "ts-loader": "^9.2.5",
+    "webpack": "^5.52.1",
+    "webpack-cli": "^4.8.0",
+    "@vscode/test-electron": "^1.6.2"
+  },
+  "dependencies": {
+    "@sap-devx/npm-dependencies-validation": "file:../npm-dependencies-validation"
```
bd82, line 52: the mono-repo should take care of this.
you should not need to use a **relative** file reference

.vscode/launch.json
```diff
@@ -41,6 +41,18 @@
       "outFiles": [
         "${workspaceFolder}/packages/app-studio-toolkit/dist/**/*.js"
       ]
+    },
+    {
+      "name": "Run Validation",
```
bd82, line 46: Use the **exact** name of the vscode extension.
This is because we have many launch configurations here so we should start
to be more strict.

packages/npm-dependencies-validation/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@sap-devx/npm-dependencies-validation",
+  "version": "1.3.0",
```
bd82, line 3: I think the mono repo is not at version `1.3.2` please align

packages/npm-dependencies-validation/package.json
bd82: This comment is about the issue of api.ts (`export {x,y,z} from "..."`) we discussed in voice.

packages/npm-dependencies-validation/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "main": "./dist/src/api",
+  "types": "./dist/src/index.d.ts",
+  "scripts": {
+    "test": "mocha",
+    "compile:watch": "tsc -p ./ -watch",
+    "compile": "tsc -p ./",
+    "coverage": "nyc mocha"
+  },
+  "devDependencies": {
+    "@types/vscode": "^1.62.0"
```
bd82, line 21: used a fixed version

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
+import { spawnCommand } from "./utils/npmUtil";
```
bd82, line 1: can we rename this file to something more meaningful?
perhaps named after the single function it exports?

bd82: imports should be ordered according to these rules
- https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  VscodeFsUri,
+} from "./types";
+
+const LS_DEPS = ["ls", "--depth=0"]; // dependencies and extraneous modules
+const LS_DEV_DEPS = [...LS_DEPS, "--dev"]; // devDependencies
+
+async function getDepsStatus(
+  packageJsonPath: string,
+  devDeps: boolean
+): Promise<any> {
```
bd82, line 17: try to define the type that you are returning

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  DependencyIssue,
+  IssueType,
+  NpmLsDependency,
+  VscodeFsUri,
+} from "./types";
+
+const LS_DEPS = ["ls", "--depth=0"]; // dependencies and extraneous modules
+const LS_DEV_DEPS = [...LS_DEPS, "--dev"]; // devDependencies
+
+async function getDepsStatus(
```
bd82, line 14: `getDepsStatus` name may not be the most optimal.

- `get` prefix is misleading, its not a "standard" getter.
- also its not descriptive enough, other options:
  - `listNodeModulesDirectDepedencies` (maybe too long).
  - `listNodeModulesDeps` (I vote for this as npm uses the `list` verb too).
  - `readNodeModuleDeps` (not sure about `read` verb).

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  packageJsonPath: string,
+  devDeps: boolean
+): Promise<any> {
+  // try to use DependencyType ???
+  return spawnCommand(
```
bd82, line 19: We will be executing more npm commands in the near future (`prune / install`).
And will have additional shared concerns while executing them:
- output redirection
- timeout
- ?

Perhaps there should be an `npmCommandsInvoker.ts` file to handle this concern.
This could also make it easy to refactor the implementation in the future, e.g if we switch to the npm's programatic APi.

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    devDeps ? [...LS_DEV_DEPS] : [...LS_DEPS],
+    resolve(dirname(packageJsonPath))
+  );
+}
+
+function getProblemType(dependency: NpmLsDependency): IssueType | undefined {
+  if (dependency.missing) return "missing";
+  if (dependency.invalid) return "invalid";
+  if (dependency.extraneous) return "extraneous";
```
bd82, line 28: - consider adding branch for **explicit** return of undefined
- what is better here (if or if / else) ?
- can you link to the blog post which recommand this style? it looks good, and I would want to read the discussion there

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export async function getDependencyIssues(
+  uri: VscodeFsUri
+): Promise<DependencyIssue[]> {
+  const managedByNpm6 = await isManagedByNpm6(uri);
+  if (!managedByNpm6) return [];
+
+  const [depsAndRedundantIssues, devDepsIssues] = await Promise.all([
+    getIssues(uri.fsPath),
```
bd82, line 64: need to explain in-depth why two commands are needed

packages/npm-dependencies-validation/src/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  type: IssueType;
+  devDependency?: boolean;
+};
+
+export type NpmLsDependency = {
+  // TODO: try to extend to DependencyType ??
+  [key in IssueType]?: boolean;
+};
+
+// Pick ???
```
bd82, line 15: see if you can use `@types/vscode`

packages/npm-dependencies-validation/src/types.ts
```diff
@@ -0,0 +1,22 @@
+export type IssueType = "missing" | "invalid" | "extraneous";
+
+export type DependencyIssue = {
```
bd82, line 3: `DependencyIssue` --> `NPMDependencyIssue` ?

packages/npm-dependencies-validation/src/types.ts
```diff
@@ -0,0 +1,22 @@
+export type IssueType = "missing" | "invalid" | "extraneous";
```
bd82, line 1: `IssueType` --> `NPMIssueType`

Note these types are exposed externally (via `getDependencyIssues`) so good name are important

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      if (type !== "extraneous") {
+        problematicDep.devDependency = devDeps;
+      }
+      problematicDeps.push(problematicDep);
+    }
+  }
+  return problematicDeps;
+}
+
+export async function getDependencyIssues(
```
bd82, line 57: `get` --> `compute` / `calculate` ?

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  );
+}
+
+function getProblemType(dependency: NpmLsDependency): IssueType | undefined {
+  if (dependency.missing) return "missing";
+  if (dependency.invalid) return "invalid";
+  if (dependency.extraneous) return "extraneous";
+}
+
+async function getIssues(
```
bd82, line 31: `get` -> `compute` / `calculate` ?

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  path: string,
+  devDeps = false
```
bd82, line 33: - config object?
- I don't think we want a default value for `devDeps` as it can add a branch that needs testing for 100% coverage.
-

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+): Promise<DependencyIssue[]> {
+  const { dependencies } = await getDepsStatus(path, devDeps);
+
+  const problematicDeps: DependencyIssue[] = [];
+  for (const name in dependencies) {
+    const dependency = dependencies[name];
+    const type = getProblemType(dependency);
+    if (type) {
+      const version = dependency.version ?? dependency.required;
```
bd82, line 42: For missing deps should not we have **two** versions?
meaning actual vs needed?

Do we need those two versions to provide a good error message?

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+async function getIssues(
+  path: string,
+  devDeps = false
+): Promise<DependencyIssue[]> {
+  const { dependencies } = await getDepsStatus(path, devDeps);
+
+  const problematicDeps: DependencyIssue[] = [];
+  for (const name in dependencies) {
```
bd82, line 38: seems like you are combining a **map** and **filter**.
try to do it in **two steps** using functional logic (lodash).

1. filter to only keep the deps that have issues
2. map those problemDeps to `NPMDepsIssues`

Basically when reading the code one is less interested in **how** you do it (imperative style).
but **what** you are trying to acomplish.

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    const dependency = dependencies[name];
+    const type = getProblemType(dependency);
+    if (type) {
+      const version = dependency.version ?? dependency.required;
+      const problematicDep: DependencyIssue = {
+        name,
+        version,
+        type,
+      };
+      if (type !== "extraneous") {
```
bd82, line 48: why is this branch needed (dev + extraneous)?

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      problematicDeps.push(problematicDep);
+    }
+  }
+  return problematicDeps;
+}
+
+export async function getDependencyIssues(
+  uri: VscodeFsUri
+): Promise<DependencyIssue[]> {
+  const managedByNpm6 = await isManagedByNpm6(uri);
```
bd82, line 60: `isManagedByNpm6` --> `isSupportedPackageJson` 

The reason is that npm6 won't exist in most workspaces in ~1.5 years.
and in the future we may support more and more variants.

bd82: I wonder if somebody in the npm ecosystem implemented the logic to determine "isYarn" / "isPnpm" / ...

packages/npm-dependencies-validation/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (!managedByNpm6) return [];
+
+  const [depsAndRedundantIssues, devDepsIssues] = await Promise.all([
+    getIssues(uri.fsPath),
+    getIssues(uri.fsPath, true), // config object ???
+  ]);
+  return [...depsAndRedundantIssues, ...devDepsIssues];
```
bd82, line 67: we are merging the issues of two `npm ls` commands rresults?
Are we certain the merged result will never contain duplicates?

If so please comment and explain why in the code.

packages/npm-dependencies-validation/src/index.ts
bd82: comment and explain the `--depth=0` argument (direct deps only)

packages/npm-dependencies-validation/src/api.ts
```diff
@@ -0,0 +1 @@
+export { getDependencyIssues } from "./index";
```
bd82, line 1: you may also want to export relevant types here (e.g used in the `getDependencyIssues` signature)

packages/npm-dependencies-validation/src/index.ts
bd82: nevermind i thought the spawn is here directly.

packages/npm-dependencies-validation/src/utils/npmUtil.ts
```diff
@@ -0,0 +1,26 @@
+import { spawn } from "child_process";
+
+export function getNPM(): string {
+  return /^win/.test(process.platform) ? "npm.cmd" : "npm";
+}
+
+export function spawnCommand(
```
bd82, line 7: `SpawnCommand` name is not the most descriptive.

- It is **implemntation** details that we use `spawn`
 - maybe we will swith to `exec`?
 - maybe we will switch to programatic api?

Maybe `SpawnCommand` --> `invokeNPMCommand` ?
- can replace `invoke` with `run` / `perform` / ???

packages/npm-dependencies-validation/src/utils/npmUtil.ts
```diff
@@ -0,0 +1,26 @@
...
+  commandArgs: string[],
+  workingDir: string
```
bd82, line 9: no need for two terms to describe the same things `cwd` vs `workingDir`

packages/npm-dependencies-validation/src/utils/npmUtil.ts
bd82: although, maybe because we expose `  commandArgs: string[],` param
we can't really **abstract* away that a child process is run here.

packages/npm-dependencies-validation/src/utils/npmUtil.ts
```diff
@@ -0,0 +1,26 @@
...
+): Promise<any> {
```
bd82, line 10: consider replacing `any` with generic `T` so the invoker can supply the expected return type.
also consider making three variants (ls/prune/install) so we will have an exact return type.

- If you make three variants we can abstract away the general `commandArgs`

packages/npm-dependencies-validation/src/utils/npmUtil.ts
```diff
@@ -0,0 +1,26 @@
...
+  return new Promise((resolve, reject) => {
+    const command = spawn(getNPM(), [...commandArgs, "--json"], {
+      cwd: workingDir,
+    });
+
+    command.stdout.on("data", (data) => {
```
bd82, line 16: is there any max size buffer here?
does all the JSON gets passed in a **single** `data` event?
Please verify using the node docs

packages/npm-dependencies-validation/src/utils/npmUtil.ts
```diff
@@ -0,0 +1,26 @@
...
+export function getNPM(): string {
+  return /^win/.test(process.platform) ? "npm.cmd" : "npm";
+}
+
+export function spawnCommand(
+  commandArgs: string[],
+  workingDir: string
+): Promise<any> {
+  return new Promise((resolve, reject) => {
+    const command = spawn(getNPM(), [...commandArgs, "--json"], {
```
bd82, line 12: note the `"--json"` already assumes this is an `invokeNPMListCommand` rather than a **general** invoke npm command. 

So maybe continue this trend make the three variants which abstract away the specifics of each command

bd82: - is there a promise version of `spawn`?
- are there 3rd party libraries that perform this wrapping?

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -0,0 +1,116 @@
...
+    const content: { name: string } = JSON.parse(packageJsonContent);
+    return content;
+  } catch (error: any) {
+    return {};
+  }
+}
+
+async function isPathExist(packageJsonPath: string): Promise<boolean> {
+  try {
+    await access(packageJsonPath, constants.R_OK | constants.W_OK);
```
bd82, line 33: access checks the permissions, in theory our process may not have permissions
even though the file exists.
- https://nodejs.org/api/fs.html#fspromisesaccesspath-mode

bd82: I think you should use one of the fs.stat variants

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -0,0 +1,116 @@
+import { access, readFile } from "fs/promises";
+import { constants } from "fs";
+import { join, dirname, sep, relative } from "path";
+import { VscodeFsUri, VscodeWsFolder } from "../types";
+
+const yarnManagerFiles = [
```
bd82, line 6: we need to decide exactly which files to test for.

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -0,0 +1,116 @@
...
+  ".yarn",
+  "./node_modules/.yarn-integrity",
+];
+const pnpmManagerFiles = [
+  "pnpm-workspace.yaml",
+  "pnpm-lock.yaml",
+  ".pnpmfile.cjs",
+  "./node_modules/.pnpm",
+];
+const npm7Properties = ["workspaces"];
```
bd82, line 19: is this `npm7props` or `monorepoProps`?

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -0,0 +1,116 @@
...
+
+function isManagedByPnpm(packageJsonFileUri: VscodeFsUri): Promise<boolean> {
+  return hasNonNPMManagerFiles(pnpmManagerFiles, packageJsonFileUri);
+}
+
+async function isMonoRepoRoot(
+  packageJsonFileUri: VscodeFsUri
+): Promise<boolean> {
+  const content = await readJsonFile(packageJsonFileUri);
+  return npm7Properties.some((property) => {
```
bd82, line 99: use lodash's `some`.

because while `Array.some` is supported in all modern nodejs versions.
the habit of using built-in array methods can cause issues:
- Not everything built-in JavaScript is safe to use on **all** node versions.
- We eventually run into some method that we need lodash for, and lodash style is different than native method style so we have inconsistent code base.
  - [].some(()  => { }) vs some([], () => { })

bd82: not sure. but the traversing on all the props may be too generic.
this depends if we have ever additional props we need to check for.

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -0,0 +1,116 @@
...
+    return property in content;
+  });
+}
+
+export async function isManagedByNpm6(
```
bd82, line 104: I know we discussed this name already and sorry for confusing you 😢 

After more thought I now think something like:
- `isSupportedPackageManager` 
- `isCurrentlySupported` (maybe most accurate because, it is not just about the pkg manager).

The reason is that while we now support only npm6, in the future we will certinally support npm7, and maybe some versions of workspaces (mono-repos). so then the function name will have to be adjusted.

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -0,0 +1,116 @@
...
+async function isPathExist(packageJsonPath: string): Promise<boolean> {
+  try {
+    await access(packageJsonPath, constants.R_OK | constants.W_OK);
+    return true;
+  } catch (error) {
+    return false;
+  }
+}
+
+// async function findParentPackageJsonPath(
```
bd82, line 40: remove commented out code, (it remains in history of PR if u will need it

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -0,0 +1,116 @@
...
+  try {
+    const packageJsonContent = await readFile(packageJsonUri.fsPath, "utf-8");
+    const content: { name: string } = JSON.parse(packageJsonContent);
+    return content;
+  } catch (error: any) {
+    return {};
+  }
+}
+
+async function isPathExist(packageJsonPath: string): Promise<boolean> {
```
bd82, line 31: `isPathExist` --> `isFileExist`?

bd82: `packageJsonPath` arg --> absPath (assuming it must be absolute)

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -0,0 +1,116 @@
...
+//       (await readJsonFile(candidatePackageJson))
+//     ) {
+//       break;
+//     }
+//   }
+
+//   return candidate ?? location;
+// }
+
+async function hasNonNPMManagerFiles(
```
bd82, line 73: `hasNonNPMManagerFiles` --> `pathContainsAnyOfFiles`

Also change `packageJsonFileUri` to simple `absPath: string` it does not need to be aware of anything related to vscode so deep into our **pure** logic

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -0,0 +1,116 @@
...
+  fileNames: string[],
+  packageJsonFileUri: VscodeFsUri
+): Promise<boolean> {
+  for (const fileName of fileNames) {
+    const packageJsonDirPath = join(dirname(packageJsonFileUri.fsPath));
+    if (await isPathExist(join(packageJsonDirPath, fileName))) {
```
bd82, line 79: need to investiate ability to Promise.All but exist early on first resolution to true

bd82: There is Promise.any but it is not supported in nodejs 14
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any

packages/npm-dependencies-validation/test/index.spec.ts
```diff
@@ -0,0 +1,98 @@
+import { getDependencyIssues } from "../src/index";
```
bd82, line 1: import order 
- https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

packages/vscode-dependencies-validation/webpack.config.js
```diff
@@ -0,0 +1,48 @@
+//@ts-check
```
bd82, line 1: dont use a new generated webpack config.
build upon the existing one in app-studio-toolkit for consistency.

packages/vscode-dependencies-validation/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "dist",
+    "baseUrl": ".",
+    "skipLibCheck": true
```
bd82, line 7: why     "skipLibCheck": true ?

packages/vscode-dependencies-validation/README.md
```diff
@@ -0,0 +1,71 @@
+# vscode-dependencies-validation README
```
bd82, line 1: empty out generated text here...

packages/vscode-dependencies-validation/.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
bd82, line 1: we use the root launch.json
maybe remove this?
also add any none root `.vscode` to .gitignore?

packages/vscode-dependencies-validation/.mocharc.js
```diff
@@ -0,0 +1,8 @@
+const chai = require("chai");
+const chaiAsPromised = require("chai-as-promised");
```
bd82, line 2: evaluate option of having a root configuration for mocha.
and maybe customizing (extending it) per specific sub-package (if needed).

packages/vscode-dependencies-validation/.eslintignore
```diff
@@ -0,0 +1,6 @@
+node_modules
```
bd82, line 1: please evaluate option for extacting common eslint ignore to settings to root of mono repo

packages/npm-dependencies-validation/test/projects/no_deps_installed/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "no_deps_installed",
+  "version": "1.0.0",
+  "dependencies": {
+    "lodash": "~4.17.21",
```
bd82, line 5: use fixed version as it may change test outputs in future if a different version is installed

packages/npm-dependencies-validation/test/projects/no_deps_installed/package.json
```diff
@@ -0,0 +1,11 @@
...
+    "json-fixer": "1.6.12"
+  },
+  "devDependencies": {
+    "typescript": "^4.4.4"
```
bd82, line 9: fixed versions

packages/npm-dependencies-validation/test/projects/some_deps_installed/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "no_deps_installed",
+  "version": "1.0.0",
+  "dependencies": {
+    "json-fixer": "^1.6.12",
```
bd82, line 5: fixed versions

packages/npm-dependencies-validation/test/index.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+import { expect, assert } from "chai";
+import * as fsPromises from "fs/promises";
+import { SinonMock, createSandbox } from "sinon";
+
+describe("index unit test", () => {
```
bd82, line 6: the describe/context/it should be written in BDD style
- https://mochajs.org/#bdd
- https://en.wikipedia.org/wiki/Behavior-driven_development#Specification_as_a_ubiquitous_language (different DSL verbs, but same concepts)

packages/npm-dependencies-validation/test/index.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  it("2 dependencies and 1 devDependency declared but are not installed", async () => {
+    const result = await getDependencyIssues({
+      fsPath: "./test/projects/no_deps_installed/package.json",
+    });
+    expect(result).to.have.lengthOf(3);
```
bd82, line 11: evaluate use of https://www.chaijs.com/plugins/deep-equal-in-any-order/

Not sure its optimal, may be less verbose than the current implementation, however also harder to debug?

packages/npm-dependencies-validation/test/index.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    // redundant json-fixer devDependency
+    const jsonFixer = result.find((dep) => dep.name === "json-fixer");
+    assert.isDefined(jsonFixer);
+    expect(jsonFixer?.devDependency).to.be.undefined;
+    expect(jsonFixer?.type).to.be.equal("extraneous");
+    expect(jsonFixer?.version).to.be.equal("1.6.12");
+  });
+
+  it("invalid package.json", async () => {
+    const fsMock: SinonMock = createSandbox().mock(fsPromises);
```
bd82, line 77: mock has to be in "before" / "after" block otherwise it is not guaranteed to run after test has failed because failed expect would throw an exception so the **next** test may fail because the mock has not been restored.

packages/npm-dependencies-validation/test/index.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  it("2 dependencies and 1 devDependency declared but are not installed", async () => {
+    const result = await getDependencyIssues({
+      fsPath: "./test/projects/no_deps_installed/package.json",
+    });
+    expect(result).to.have.lengthOf(3);
+    // missing json-fixer dependency
+    const jsonFixer = result.find((dep) => dep.name === "json-fixer");
+    assert.isDefined(jsonFixer);
+    expect(jsonFixer?.devDependency).to.be.false;
+    expect(jsonFixer?.type).to.be.equal("missing");
```
bd82, line 16: The connection words should create proper English sentences.

`expect(jsonFixer?.type).to.be.equal("missing");` 

should instead be 
`expect(jsonFixer?.type).to.equal("missing");` 

"expect bamba to equal osem"
"expect bisli to be undefined"

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it("spawnCommand succeeded with ls", () => {
+    return expect(spawnCommand(["ls", "--depth=0"], "./")).to.be.fulfilled;
+  });
+
+  describe("npm command", () => {
+    const originalPlatform = process.platform;
+
+    after(function () {
```
bd82, line 18: after**Each**

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+describe("npmUtil unit test", () => {
+  it("spawnCommand failed", () => {
+    return expect(
+      spawnCommand(["ls"], "./non_existing_path")
+    ).to.be.rejectedWith(`spawn ${getNPM()} ENOENT`);
+  });
+
+  it("spawnCommand succeeded with ls", () => {
+    return expect(spawnCommand(["ls", "--depth=0"], "./")).to.be.fulfilled;
```
bd82, line 12: can you add a minimal assertion on its results?

bd82: should not chai as promised be used with `.eventually` keyword?
- https://github.com/domenic/chai-as-promised

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { spawnCommand, getNPM } from "../../src/utils/npmUtil";
+import { expect } from "chai";
+
+describe("npmUtil unit test", () => {
+  it("spawnCommand failed", () => {
+    return expect(
+      spawnCommand(["ls"], "./non_existing_path")
+    ).to.be.rejectedWith(`spawn ${getNPM()} ENOENT`);
```
bd82, line 8: should not chai as promised be used with .eventually keyword?

https://github.com/domenic/chai-as-promised

packages/npm-dependencies-validation/test/utils/packageJsonUtil/packageJsonUtil.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { expect } from "chai";
+import { VscodeFsUri, VscodeWsFolder } from "../../../src/types";
+import { isManagedByNpm6 } from "../../../src/utils/packageJsonUtil";
+
+describe("packageJsonUtil unit test", () => {
+  it("not manged by npm, has pnpm-workspace.yaml", async () => {
```
bd82, line 6: add a test when it returns true

packages/npm-dependencies-validation/test/utils/packageJsonUtil/packageJsonUtil.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    const uri: VscodeFsUri = {
+      fsPath:
+        "test/utils/packageJsonUtil/projects/pnpm-workspace/wsFolder1/child/package.json",
```
bd82, line 9: consider placing the "samples" closer to the tests (e.g in sibling folder)
instead of one single "samples" (projects) folder.
it makes it harder to understand which samples are redundant.

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -0,0 +1,51 @@
+import { getDependencyIssues } from "@sap-devx/npm-dependencies-validation";
+
+import { DependencyIssue } from "@sap-devx/npm-dependencies-validation/dist/src/types";
+
+import { Uri, ExtensionContext, workspace, window } from "vscode";
```
bd82, line 5: import order (monorepo deps **after** external)

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -0,0 +1,51 @@
...
+    void displayProblematicDependencies(uri);
+  });
+
+  fileWatcher.onDidCreate((uri: Uri) => {
+    void displayProblematicDependencies(uri);
+  });
+
+  fileWatcher.onDidDelete((uri: Uri) => {
+    //TODO: check if we need it ??
+    void displayProblematicDependencies(uri);
```
bd82, line 35: check what happens when a file with diagnostics errors get deleted in vscode

packages/npm-dependencies-validation/test/index.spec.ts
bd82: This comment is for **all** the spec files

packages/npm-dependencies-validation/test/index.spec.ts
```diff
@@ -0,0 +1,98 @@
+import { getDependencyIssues } from "../src/index";
+import { expect, assert } from "chai";
+import * as fsPromises from "fs/promises";
+import { SinonMock, createSandbox } from "sinon";
+
+describe("index unit test", () => {
+  it("2 dependencies and 1 devDependency declared but are not installed", async () => {
+    const result = await getDependencyIssues({
+      fsPath: "./test/projects/no_deps_installed/package.json",
```
bd82, line 9: evaluate some manner to link project (samples) files with test file.
otherwise overtime redundant samples will get collected

packages/npm-dependencies-validation/test/index.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    expect(lodash?.version).to.be.equal("~4.17.21");
+    // missing typescript devDependency
+    const typescript = result.find((dep) => dep.name === "typescript");
+    assert.isDefined(typescript);
+    expect(typescript?.devDependency).to.be.true;
+    expect(typescript?.type).to.be.equal("missing");
+    expect(typescript?.version).to.be.equal("^4.4.4");
+  });
+
+  it("1 dependency and 1 devDependency declared but are not installed, 1 dependency is installed", async () => {
```
bd82, line 32: howcome is that one dep installed?
who calls `npm install` to make sure it is installed?

packages/npm-dependencies-validation/test/index.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    // missing typescript devDependency
+    const typescript = result.find((dep) => dep.name === "typescript");
+    assert.isDefined(typescript);
+    expect(typescript?.devDependency).to.be.true;
+    expect(typescript?.type).to.be.equal("missing");
+    expect(typescript?.version).to.be.equal("^4.4.4");
+  });
+
+  it("1 dependency and 1 devDependency declared but not installed, 1 dependency is installed but redundant", async () => {
+    const result = await getDependencyIssues({
```
bd82, line 52: I dont see any automation to ensure some of the deps would get installed
how does this work?

packages/npm-dependencies-validation/.mocharc.js
slavik-lvovsky: It depends. On my machine sometimes some tests took 4000+ millis, but sometimes the same tests took less than 2000 millis. I think it is OK to leave 5000 millis for all tests.

.vscode/launch.json
slavik-lvovsky: Done

packages/npm-dependencies-validation/.npmignore
slavik-lvovsky: Done

packages/npm-dependencies-validation/package.json
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/api.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done
It was a video on Youtube in russian :)

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: dependencies is not an array it is an object. Like this:
"dependencies": {
    "typescript": {
      "required": "4.4.4",
      "missing": true
    }
  }

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: A dependency is missing when it is in package.json but not installed. There are no 2 versions in this case

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Removed.

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done.
I will try to find

packages/npm-dependencies-validation/src/types.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/types.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/types.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/utils/npmUtil.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/utils/npmUtil.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/utils/npmUtil.ts
slavik-lvovsky: Done.
Why do we need three variants when the only difference is command args ?

packages/npm-dependencies-validation/src/utils/npmUtil.ts
slavik-lvovsky: there is no official async version of spawn.
One 3rd party lib , that I found, is not looking good

packages/npm-dependencies-validation/src/utils/npmUtil.ts
slavik-lvovsky: By default, pipes for stdin, stdout, and stderr are established between the parent Node.js process and the spawned subprocess. These pipes have limited (and platform-specific) capacity. If the subprocess writes to stdout in excess of that limit without the output being captured, the subprocess blocks waiting for the pipe buffer to accept more data.

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: Done. monorepoProps

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: Why not all of them ?

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: It is used not just for files.
Changed to absPath

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: Using fs.stat() to check for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Instead, user code should open/read/write the file directly and handle the error raised if the file is not available.

To check if a file exists without manipulating it afterwards, fs.access() is recommended.

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: I do not think Promise.all could improve something here

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: Why do we need to use here lodash if array.some is supported in node 14?

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/test/index.spec.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/package.json
slavik-lvovsky: Done

packages/npm-dependencies-validation/package.json
slavik-lvovsky: Done

packages/npm-dependencies-validation/package.json
slavik-lvovsky: Done

packages/vscode-dependencies-validation/tsconfig.json
slavik-lvovsky: We have it in app-studio-toolkit.
Removed.

packages/vscode-dependencies-validation/src/extension.ts
slavik-lvovsky: it is removed from problems view
I will check whether we need to handle this type of action

packages/vscode-dependencies-validation/src/extension.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/package.json
slavik-lvovsky: Done

packages/vscode-dependencies-validation/README.md
slavik-lvovsky: Done

packages/vscode-dependencies-validation/.vscode/launch.json
slavik-lvovsky: Removed.
What do you mean by "none root .vscode" ?

packages/npm-dependencies-validation/test/projects/no_deps_installed/package.json
slavik-lvovsky: Nothing is installed in tests.
I want to check that all types of versions are returned correctly

packages/npm-dependencies-validation/test/projects/no_deps_installed/package.json
slavik-lvovsky: Needed by test

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
slavik-lvovsky: No, I wanted it to set original value only after all tests in this describe

packages/vscode-dependencies-validation/package.json
slavik-lvovsky: Done

packages/vscode-dependencies-validation/webpack.config.js
slavik-lvovsky: Done

packages/npm-dependencies-validation/test/index.spec.ts
slavik-lvovsky: node_modules has the relevant package. See the folder path.
I get the right result when running ls --depth=0 --json on the test folder

packages/npm-dependencies-validation/test/index.spec.ts
slavik-lvovsky: Not sure what do you mean.
The test has relevant description and path to the tested folder

packages/npm-dependencies-validation/test/utils/packageJsonUtil/packageJsonUtil.spec.ts
slavik-lvovsky: the samples are under specific test folder: test/utils/packageJsonUtil/projects

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
slavik-lvovsky: Here I just wanted to check that the command succeeded.
.be.fulfilled is also part of the chai-as-promised API

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
slavik-lvovsky: not always, this one is also OK

packages/npm-dependencies-validation/test/index.spec.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/test/index.spec.ts
slavik-lvovsky: I am not sure it is necessary for current tests.
Let's talk about it

packages/npm-dependencies-validation/test/index.spec.ts
slavik-lvovsky: What exactly wrong in tests ? 
Let's talk about it

packages/npm-dependencies-validation/src/index.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/.mocharc.js
slavik-lvovsky: Let's talk about it

bd82:
merging this into the https://github.com/SAP/app-studio-toolkit/tree/main_versions_misalignment and will respond to CR fixes and comments async

packages/npm-dependencies-validation/.mocharc.js
bd82: normally it is not good to have a high timeout for **all** tests.
because if there is a general issue and everything fails on timeout it can take a long while to get feedback from the tests.
Not sure if that is relevant here (what percentage of the tests are long running.

packages/npm-dependencies-validation/src/index.ts
bd82: :)

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
bd82: its okay, although exit code "0" is a weak assertion

packages/npm-dependencies-validation/test/utils/packageJsonUtil/packageJsonUtil.spec.ts
bd82: for utilities it can be more thorough to do `.only` on current spec file and ensure 100% coverage for speciic tested file
via **only** it's matching ".spec" file.

But it is not mandatory

packages/npm-dependencies-validation/test/utils/packageJsonUtil/packageJsonUtil.spec.ts
bd82: I meant a test sub-folder per test
so if we delete a test in the future we will also know which sub-folder to delete.

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
bd82: this damages tests isolation.
the Nth test has a value from the Nth-1 test...
So execution could be different, e.g if we run all the tests in sequence or if we run only one test.
(I wasted hours on a similar problem in this very same repo...)

packages/npm-dependencies-validation/test/index.spec.ts
bd82: i will try to implement this

packages/npm-dependencies-validation/test/index.spec.ts
bd82: will checkout the code to examine this

packages/npm-dependencies-validation/test/index.spec.ts
bd82: will checkout to understand this

packages/npm-dependencies-validation/test/index.spec.ts
bd82: that was the issue with the .gitignore we discovered today

packages/npm-dependencies-validation/test/index.spec.ts
bd82: this was the issue with the .gitignore

PR #154, opened 2021-12-02
feat: simplified solution versions misalignment

slavik-lvovsky:


packages/npm-dependencies-validation/package.json
```diff
@@ -24,5 +24,8 @@
     "clean": "rimraf ./dist",
     "test": "mocha",
     "ci": "npm-run-all clean compile coverage"
+  },
+  "dependencies": {
+    "@types/vscode": "^1.50.0"
```
bd82, line 29: npm-dep-validation should not depend on vscode (even types).
The logging/outputChannel dep should be in a more generic fashion
e.g a generic logger interface that the consumer (our vscode ext) implements.

packages/vscode-dependencies-validation/tsconfig.json
```diff
@@ -5,5 +5,5 @@
     "outDir": "dist",
     "baseUrl": "."
   },
-  "include": ["src/**/*", "test/**/*"]
+  "include": ["src/**/*", "test/**/*", "test"]
```
bd82, line 8: why is this needed?

packages/npm-dependencies-validation/src/depIssues.ts
```diff
@@ -0,0 +1,30 @@
...
+  const commandConfig = {
+    commandArgs: [...LS_ARGS],
+    cwd: dirname(absPackageJsonPath),
+  };
+  const npmLsResult = await invokeNPMCommand<NpmLsResult>(commandConfig);
+  const problemsLength = npmLsResult.problems?.length || 0;
+  if (problemsLength != 0) return npmLsResult;
+
+  // devDependencies issues and extraneous modules
+  commandConfig.commandArgs.push("--dev");
```
bd82, line 28: avoid **mutating** the `commandArgs` this could make debugging more complicated
because someone could assume the first command was also executed with the "--dev".

Instead clone the object and assign to a new variable and only modify the new variable.
- this new variable should also have a descriptive name to distinguish it from the previous one.

packages/npm-dependencies-validation/src/depIssues.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!currentlySupported) return { problems: [] };
+
+  // dependencies issues and extraneous modules
+  const commandConfig = {
+    commandArgs: [...LS_ARGS],
+    cwd: dirname(absPackageJsonPath),
+  };
+  const npmLsResult = await invokeNPMCommand<NpmLsResult>(commandConfig);
+  const problemsLength = npmLsResult.problems?.length || 0;
+  if (problemsLength != 0) return npmLsResult;
```
bd82, line 25:
```typescript
const problemsLength = npmLsResult.problems?.length || 0;
  if (problemsLength != 0) return npmLsResult;
```

This looks a bit convoluted for the purpose of checking `isEmpty(npmLsResult.problems)`
Maybe you can replace it with `isEmpty` check?

bd82: also "!=" should be "!=="

bd82:
```typescript
return npmLsResult;
```

This early return has a-lot of meaning.
We are ignoring devDeps and extraneous dependencies **if** there are other issues.
So an end user would have to perform the "npm install" **twice** in some edge cases.
And could be notified of issues twice in one flow.

This assumption / behavior needs to be :
- Well documented in the source code
- Synced with Product
- Synced with head arch

packages/npm-dependencies-validation/src/types.ts
```diff
@@ -1,30 +1,27 @@
-export type NPMIssueType = "missing" | "invalid" | "extraneous";
+import { OutputChannel } from "vscode";
```
bd82, line 1: as mentioned else where.
we can avoid a dependency to vscode types by declaring our own mini logger interface.
and wrapping the output channel in a suitable way when these APIs are called from a VSCode extension.

packages/npm-dependencies-validation/src/types.ts
```diff
@@ -1,30 +1,27 @@
...
 
-export type Version = "required" | "version";
-
-export type NPMDependencyIssue = {
-  name: string;
+export type NpmLsResult = {
+  name?: string;
```
bd82, line 4: `name` is optional ?

packages/npm-dependencies-validation/src/types.ts
```diff
@@ -1,30 +1,27 @@
...
+export type PackageJson = {
   name: string;
   version: string;
-  dependencies: Dependencies;
-};
+} & DependenciesProperties;
 
-export type Dependencies = {
-  [key: string]: Dependency;
+export type DependenciesProperties = {
```
bd82, line 16: This seems a bit complex, could we just declare everything directly on PackageJson type?
I could not find where else are using these sub-types that are used to assemble `PackageJson `

packages/npm-dependencies-validation/src/utils/npmUtil.ts
```diff
@@ -1,23 +1,46 @@
...
 
 export function getNPM(): string {
   return /^win/.test(process.platform) ? "npm.cmd" : "npm";
 }
 
 export function invokeNPMCommand<T>(
-  commandArgs: string[],
-  cwd: string
+  config: NpmCommandConfig,
+  outputChannel?: VscodeOutputChannel
```
bd82, line 10: this code should not be aware of the concept of "outputChannel"

bd82: There are two levels of logging here (log vs error)
But no log level, either both are logged or none.
Another reason to wrap in a mini logger interface...

packages/npm-dependencies-validation/src/utils/npmUtil.ts
```diff
@@ -1,23 +1,46 @@
...
     command.on("error", (error) => {
+      showError(error, outputChannel);
       reject(error);
     });
   });
 }
+
+function showError(error: Error, outputChannel?: VscodeOutputChannel): void {
+  const { stack } = error;
+  console.error(stack);
```
bd82, line 31: **duplicating** console.log/error calls and sending to vscode outputChannel 
is another example why we should extract the logger interface.

We naturally do not want to always log to the console...

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -17,17 +18,18 @@ const pnpmManagerFiles = [
...
-async function readJsonFile(jsonFilePath: string): Promise<any> {
+async function readJsonFile(jsonFilePath: string): Promise<PackageJson> {
   try {
     const packageJsonContent = await readFile(jsonFilePath, "utf-8");
-    const content: { name: string } = JSON.parse(packageJsonContent);
+    const content: PackageJson = JSON.parse(packageJsonContent);
     return content;
   } catch (error) {
-    return {};
+    // TODO: should we return null/undefined ???
```
bd82, line 27: should we log something in the catch block?

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -17,17 +18,18 @@ const pnpmManagerFiles = [
 ];
 const monorepoProps = ["workspaces"];
 
-async function readJsonFile(jsonFilePath: string): Promise<any> {
+async function readJsonFile(jsonFilePath: string): Promise<PackageJson> {
```
bd82, line 21: do we use fs-extra here?
I think it has a readJson utility method

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -17,17 +18,18 @@ const pnpmManagerFiles = [
...
     return content;
   } catch (error) {
-    return {};
+    // TODO: should we return null/undefined ???
+    return {} as PackageJson;
   }
 }
 
-async function isPathExist(absPath: string): Promise<boolean> {
+export async function isPathExist(absPath: string): Promise<boolean> {
```
bd82, line 32: I think I already commented about the usage of `access` to check if a file exists.
Not sure it is the most suitable way (vs `stat`) also why would we care if there is a **WRITE** permission to the file?
we only ever **READ** the file in these flows.

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
```diff
@@ -41,10 +43,7 @@ async function pathContainsAnyOfFiles(
   absPath: string
 ): Promise<boolean> {
   for (const fileName of fileNames) {
-    const packageJsonDirPath = join(dirname(absPath));
-    if (await isPathExist(join(packageJsonDirPath, fileName))) {
-      return true;
-    }
+    if (await isPathExist(join(dirname(absPath), fileName))) return true;
```
bd82, line 46: document this is written with for-of to allow early exit for async code

packages/npm-dependencies-validation/test/depIssues.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { expect, assert } from "chai";
+import { resolve } from "path";
+import { findDependencyIssues } from "../src/api";
+
+describe("dependencyIssues unit test", () => {
```
bd82, line 5: please remind me to demonstrate BDD testing style descriptors next week

packages/npm-dependencies-validation/test/depIssues.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      problem.startsWith("missing: json-fixer@~1.6.12")
+    );
+    assert.isDefined(jsonFixerProblem);
+  });
+
+  it("no dependency issues are found, package.json is not found", async () => {
+    const result = await findDependencyIssues(
+      "./test/projects/non_existing_folder/package.json"
+    );
+    expect(result.problems).to.have.lengthOf(0);
```
bd82, line 42: I think there is a `to.be.empty` assertion

packages/npm-dependencies-validation/test/depIssues.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  it("no dependency issues are found, package is not supported", async () => {
+    const result = await findDependencyIssues(
+      resolve("./test/projects/not_supported/package.json")
+    );
+    expect(result.problems).to.have.lengthOf(0);
```
bd82, line 49: I think there is a `to.be.empty` assertion

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
```diff
@@ -1,15 +1,55 @@
+import { fail } from "assert";
```
bd82, line 1: using `fail` from `assert` is strange as `chai` provides it's own `assert.fail`
But I think we may not need `fail` at all

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
```diff
@@ -1,15 +1,55 @@
...
-  it("invokeNPMCommand failed", () => {
-    return expect(
-      invokeNPMCommand(["ls"], "./non_existing_path")
-    ).to.be.rejectedWith(`spawn ${getNPM()} ENOENT`);
+  const sandbox = createSandbox();
+  const outputChannel: VscodeOutputChannel = {
+    append: (data: string) => console.log(data),
+  };
+
+  afterEach(function () {
```
bd82, line 13: why is this `function ()` while others are lambda ? `() =>`

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
```diff
@@ -1,15 +1,55 @@
...
+      const config = { commandArgs: ["ls"], cwd: "./non_existing_path" };
+      return expect(invokeNPMCommand(config)).to.be.rejectedWith(
+        `spawn ${getNPM()} ENOENT`
+      );
+    });
+
+    it("outputChannel is provided", () => {
+      const appendSpy: SinonSpy = sandbox.spy(outputChannel, "append");
+      const config = { commandArgs: ["ls"], cwd: "./non_existing_path" };
+      return invokeNPMCommand(config, outputChannel)
```
bd82, line 28: it would be more **declarative and less error prone** to use `chai-as-promised` and `to.be.rejectedWith(...)`
instead of `.then / .catch`
Not sure how to combine it with the assertion on the sinonSpy, but may be worth a shot.

bd82: What happens when you do?

```typescript
async () => {
      const appendSpy: SinonSpy = sandbox.spy(outputChannel, "append");
      const config = { commandArgs: ["ls"], cwd: "./non_existing_path" };
      await invokeNPMCommand(config, outputChannel).to.be.rejectedWith(
        `spawn ${getNPM()} ENOENT`
      );
      expect(appendSpy.called).to.be.true; 
```

packages/npm-dependencies-validation/src/utils/npmUtil.ts
```diff
@@ -1,23 +1,46 @@
...
-  cwd: string
+  config: NpmCommandConfig,
+  outputChannel?: VscodeOutputChannel
 ): Promise<T> {
+  const { commandArgs, cwd } = config;
   return new Promise((resolve, reject) => {
     const command = spawn(getNPM(), [...commandArgs, "--json"], { cwd });
 
     command.stdout.on("data", (data) => {
+      showData(`${data}`, outputChannel);
```
bd82, line 17: is there a max buffer size here when we spawn the process? 
will we always get the data from JSON.parse in one big batch?

packages/npm-dependencies-validation/test/utils/packageJsonUtil/packageJsonUtil.spec.ts
```diff
@@ -29,4 +30,18 @@ describe("packageJsonUtil unit test", () => {
...
+    const packageJsonPath =
+      "test/utils/packageJsonUtil/projects/npm-managed/package.json";
+    const res = await isCurrentlySupported(packageJsonPath);
+    expect(res).to.be.true;
+  });
+
+  it("non existing package.json", async () => {
+    const packageJsonPath =
+      "test/utils/packageJsonUtil/projects/not-existing-folder/package.json";
+    const res = await isCurrentlySupported(packageJsonPath);
```
bd82, line 44: none existing path "is supported" because we only define conditions for what is NOT supported?

packages/vscode-dependencies-validation/package.json
```diff
@@ -20,13 +20,18 @@
...
     "compile": "tsc -p ./",
+    "watch": "tsc -p ./ --watch",
     "test": "mocha",
     "coverage": "nyc mocha",
     "bundle": "webpack --mode production",
     "package": "node ./scripts/package-vsix.js"
   },
   "dependencies": {
-    "@sap-devx/npm-dependencies-validation": "file:../npm-dependencies-validation"
+    "@sap-devx/npm-dependencies-validation": "file:../npm-dependencies-validation",
```
bd82, line 30: use package name, not relative link. you don't need to publish it to npm first
the pnpm/yarn/mono repo manager should handle these links

packages/vscode-dependencies-validation/package.json
```diff
@@ -20,13 +20,18 @@
     "ci": "npm-run-all clean compile bundle",
     "clean": "rimraf ./dist *.vsix",
     "compile": "tsc -p ./",
+    "watch": "tsc -p ./ --watch",
```
bd82, line 23: You are not using the top level watch?
someone could get confused and execute both the top level and sub-package levels watches...
is it worth the possible problem instead of just running the root level watch?

packages/vscode-dependencies-validation/package.json
```diff
@@ -20,13 +20,18 @@
...
     "test": "mocha",
     "coverage": "nyc mocha",
     "bundle": "webpack --mode production",
     "package": "node ./scripts/package-vsix.js"
   },
   "dependencies": {
-    "@sap-devx/npm-dependencies-validation": "file:../npm-dependencies-validation"
+    "@sap-devx/npm-dependencies-validation": "file:../npm-dependencies-validation",
+    "vfile": "4.2.1",
+    "vfile-location": "3.2.0",
```
bd82, line 32: afaik we are not using the vfile/vfile-location/jsonc-parser (yet)

packages/vscode-dependencies-validation/src/commands.ts
```diff
@@ -0,0 +1,40 @@
...
+import { invokeNPMCommand } from "@sap-devx/npm-dependencies-validation";
+import { refreshDiagnostics } from "./diagnostics";
+import { VscodeOutputChannel } from "./vscodeTypes";
+
+export async function fixAllDepIssuesCommand(
+  outputChannel: VscodeOutputChannel,
+  packageJsonPath: string,
+  dependencyIssuesDiagnosticCollection: DiagnosticCollection
+): Promise<void> {
+  outputChannel.show(true);
```
bd82, line 12: for automatic mode we should not always focus/show the output channel

packages/vscode-dependencies-validation/src/commands.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const npmCommand = "install";
+  try {
+    outputChannel.appendLine(`\nFixing dependency issues ...`);
+    const start = Date.now();
+    const config = { commandArgs: [npmCommand], cwd: dirname(packageJsonPath) };
+    await invokeNPMCommand(config, outputChannel);
```
bd82, line 19: invokeNPMCommmand prints everyting to output channel as a single JSON line.
but that is hardly human readable.
we should probably pass the outputChannel line by line instead, and not in JSON format.

packages/vscode-dependencies-validation/src/commands.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    outputChannel.append(`Done. (${Date.now() - start} millis)\n`);
+
+    void refreshPackageJsonDiagnostics(
+      packageJsonPath,
+      dependencyIssuesDiagnosticCollection
+    );
+  } catch (error) {
+    outputChannel.appendLine(`Failed: ${error.stack}`);
```
bd82, line 28: this looks like a logger level rather than an output channel level.
The logger will then re-direct it to the output channel if so configured

packages/vscode-dependencies-validation/src/commands.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+
+async function refreshPackageJsonDiagnostics(
```
bd82, line 32: should this utility function be inlined?

packages/vscode-dependencies-validation/src/commands.ts
```diff
@@ -0,0 +1,40 @@
...
+  outputChannel.show(true);
+
+  const npmCommand = "install";
+  try {
+    outputChannel.appendLine(`\nFixing dependency issues ...`);
+    const start = Date.now();
+    const config = { commandArgs: [npmCommand], cwd: dirname(packageJsonPath) };
+    await invokeNPMCommand(config, outputChannel);
+
+    outputChannel.append(`Done. (${Date.now() - start} millis)\n`);
```
bd82, line 21: I think for using facing output we would prefer to count in seconds

packages/vscode-dependencies-validation/src/diagnostics.ts
```diff
@@ -0,0 +1,42 @@
+import type { DiagnosticCollection, ExtensionContext } from "vscode";
```
bd82, line 1: 👍

packages/vscode-dependencies-validation/src/diagnostics.ts
```diff
@@ -0,0 +1,42 @@
...
+  packageJsonPath: string
+): Diagnostic {
+  const range = new Range(0, 0, 0, 10);
+  const diagnostic = new Diagnostic(
+    range,
+    problems.join("\n"),
+    0 // DiagnosticSeverity.Error
+  );
+  diagnostic.code = NPM_DEPENDENCY_ISSUES_CODE;
+  set(diagnostic, "packageJsonPath", packageJsonPath); // TODO: how to pass needed data to diagnostic ?
```
bd82, line 39: on JS objects you can always add properties.
But it is uncertain you can rely on them staying there.
For example Diagnostics in LSP are passed serialized between server and client so 
additional properties would not be serialized.

Also if diagnostics are passed between the extension-host process and the main VSCode processes these additional properties may also be lost.

What is the use case that requires an additional property here?

packages/vscode-dependencies-validation/src/diagnostics.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export async function refreshDiagnostics(
+  packageJsonPath: string,
+  dependencyIssueDiagnostics: DiagnosticCollection
+): Promise<void> {
+  const npmLsResult = await findDependencyIssues(packageJsonPath);
+
+  const diagnostics: Diagnostic[] = [];
+  if (npmLsResult.problems) {
+    diagnostics.push(
```
bd82, line 20: you don't need a  `.push` if you only ever create an array of size 1
you could create it in-place instead `[....]`

packages/vscode-dependencies-validation/src/diagnostics.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  dependencyIssueDiagnostics.set(Uri.file(packageJsonPath), diagnostics);
+}
+
+function constructDiagnostic(
+  problems: string[],
+  packageJsonPath: string
+): Diagnostic {
+  const range = new Range(0, 0, 0, 10);
```
bd82, line 32: this magical range affects where the problem will be shown to the end user (top of file)
this should be **documented**

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -1,27 +1,96 @@
...
   findDependencyIssues,
 } from "@sap-devx/npm-dependencies-validation";
+import { NPMIssuesActionProvider } from "./npmIssuesActionProvider";
+import { fixAllDepIssuesCommand } from "./commands";
+import { FIX_ALL_ISSUES_COMMAND } from "./constants";
+import { refreshDiagnostics } from "./diagnostics";
+
+const PACKAGE_JSON = "package.json";
+const PACKAGE_JSON_PATTERN = `**​/${PACKAGE_JSON}`;
+const extName = "vscode-dependency-validation";
```
bd82, line 23: is not this name (or similar) available in https://code.visualstudio.com/api/references/vscode-api#ExtensionContext
provided to the `activate` method (I think under context.extension.id)

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -1,27 +1,96 @@
...
+  NpmLsResult,
   findDependencyIssues,
 } from "@sap-devx/npm-dependencies-validation";
+import { NPMIssuesActionProvider } from "./npmIssuesActionProvider";
+import { fixAllDepIssuesCommand } from "./commands";
+import { FIX_ALL_ISSUES_COMMAND } from "./constants";
+import { refreshDiagnostics } from "./diagnostics";
+
+const PACKAGE_JSON = "package.json";
+const PACKAGE_JSON_PATTERN = `**​/${PACKAGE_JSON}`;
```
bd82, line 22: does this pattern exclude `node_modules` ?

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -1,27 +1,96 @@
...
+const extName = "vscode-dependency-validation";
 
 export function activate(context: ExtensionContext) {
+  const outputChannel = window.createOutputChannel(extName);
+  const diagnosticCollection = createDiagnosticCollection(context, extName);
+
```
bd82, line 28: - should we delay "findIssues" from startup performance considerations?
- does this scan all package.json files regardless if they are open or not? or only the open files?

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -1,27 +1,96 @@
...
+  subscribeToDocumentChanges(context, diagnosticCollection);
+
+  registerCommands(context, outputChannel, diagnosticCollection);
+}
+
+function registerCodeActionsProvider(context: ExtensionContext): void {
+  context.subscriptions.push(
+    languages.registerCodeActionsProvider(
+      { language: "json", scheme: "file", pattern: "**/package.json" },
```
bd82, line 43: did not we build this pattern already above in a cost?

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -1,27 +1,96 @@
...
+import { fixAllDepIssuesCommand } from "./commands";
+import { FIX_ALL_ISSUES_COMMAND } from "./constants";
+import { refreshDiagnostics } from "./diagnostics";
+
+const PACKAGE_JSON = "package.json";
+const PACKAGE_JSON_PATTERN = `**​/${PACKAGE_JSON}`;
+const extName = "vscode-dependency-validation";
 
 export function activate(context: ExtensionContext) {
+  const outputChannel = window.createOutputChannel(extName);
```
bd82, line 26: missing return type of activate function (void)

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -1,27 +1,96 @@
...
+  context: ExtensionContext,
+  extName: string
+): DiagnosticCollection {
+  const diagnosticCollection = languages.createDiagnosticCollection(extName);
+  context.subscriptions.push(diagnosticCollection);
+  return diagnosticCollection;
 }
 
+function registerCommands(
+  context: ExtensionContext,
```
bd82, line 62: you don't need the whole context here, just the subscriptions

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -1,27 +1,96 @@
...
   packageJsonUris.forEach((packageJsonUri) => {
     void displayProblematicDependencies(packageJsonUri);
   });
 }
 
+// TODO: somebody added yarl.lock in filesystem (not via vscode) ??
+// TODO: what should happen after git clone ??
 function addFileWatcher(): void {
-  const fileWatcher = workspace.createFileSystemWatcher("**/package.json");
+  const fileWatcher = workspace.createFileSystemWatcher(PACKAGE_JSON_PATTERN);
```
bd82, line 93: does this exclude node_modules directory?

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -1,27 +1,96 @@
...
+const PACKAGE_JSON_PATTERN = `**​/${PACKAGE_JSON}`;
+const extName = "vscode-dependency-validation";
 
 export function activate(context: ExtensionContext) {
+  const outputChannel = window.createOutputChannel(extName);
+  const diagnosticCollection = createDiagnosticCollection(context, extName);
+
   void findIssues();
 
   void addFileWatcher();
```
bd82, line 31: should the fileWatcher flow only be active in transparent flow?
were would we display diagnostics for a file which is not open?

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -1,27 +1,96 @@
...
   });
 }
 
+// TODO: somebody added yarl.lock in filesystem (not via vscode) ??
+// TODO: what should happen after git clone ??
 function addFileWatcher(): void {
-  const fileWatcher = workspace.createFileSystemWatcher("**/package.json");
+  const fileWatcher = workspace.createFileSystemWatcher(PACKAGE_JSON_PATTERN);
   fileWatcher.onDidChange((uri: Uri) => {
     void displayProblematicDependencies(uri);
```
bd82, line 95: should this be **debounced** ? we don't want to try calculating the issues on every user click...

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -41,11 +110,52 @@ async function displayProblematicDependencies(
...
 
   void window.showInformationMessage(
-    `found ${problemDeps.length} problems in ${Date.now() - start} milliseconds`
+    `found ${npmLsResult.problems?.length || 0} problems in ${
+      Date.now() - start
+    } milliseconds`
+  );
+}
+
+function subscribeToDocumentChanges(
```
bd82, line 124: this is based on vscode extension sample?

packages/vscode-dependencies-validation/src/extension.ts
```diff
@@ -41,11 +110,52 @@ async function displayProblematicDependencies(
...
+          dependencyIssueDiagnostics
+        );
+      }
+    })
+  );
+
+  context.subscriptions.push(
+    workspace.onDidChangeTextDocument(
+      (e) =>
+        void refreshDiagnostics(
```
bd82, line 149: i think this should be de-bounced for performance reasons

packages/vscode-dependencies-validation/src/npmIssuesActionProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    return action;
+  }
+}
+
+function createCommand(diagnostic: Diagnostic): Command {
+  return {
+    command: FIX_ALL_ISSUES_COMMAND,
+    title: "Fix all dependency issues",
+    arguments: [get(diagnostic, "packageJsonPath")],
```
bd82, line 52: we should check in UI5-Lang-assistanst if/how we linked commands and diagnostics.
Although that also had LSP in the middle.

packages/vscode-dependencies-validation/src/npmIssuesActionProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+    context: CodeActionContext,
+    token: CancellationToken
+  ): CodeAction[] {
+    // for each diagnostic entry create a code action command
+    return context.diagnostics
+      .filter((diagnostic) => diagnostic.code === NPM_DEPENDENCY_ISSUES_CODE)
+      .map((diagnostic) => this.createCommandCodeAction(diagnostic));
+  }
+
+  private createCommandCodeAction(diagnostic: Diagnostic): CodeAction {
```
bd82, line 34: is `createCommandCodeAction` an overridden method from `CodeActionProvider`?
if not then the name is too generic and does not describe what it does

packages/vscode-dependencies-validation/test/npmIssuesActionProvider.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { expect } from "chai";
+import * as proxyquire from "proxyquire";
+import type { CodeAction, CodeActionKind } from "vscode";
+import { NPMIssuesActionProvider } from "../src/npmIssuesActionProvider";
+import { range, textDocument, context, token, kind } from "./vscodeMocks";
+
+const diagnosticsProxy = {
```
bd82, line 7: these two proxies are only used in the single test here.
I think they should be initialized inside the describe, otherwise future tests added in this file
could cease being self contained.

packages/vscode-dependencies-validation/test/npmIssuesActionProvider.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  invokeNPMCommand() {
+    return Promise.reject("invokeNPMCommand method is not implemented");
+  },
+  "@noCallThru": true,
+};
+
+describe("npmIssuesActionProvider unit test", () => {
+  let provider: NPMIssuesActionProvider;
+
+  before(() => {
```
bd82, line 24: please wrap this **before** and **it** in another **context**
otherwise additional tests added here may not be isolated

packages/vscode-dependencies-validation/test/npmIssuesActionProvider.spec.ts
bd82: although maybe as they only implement "reject" its fine

packages/vscode-dependencies-validation/test/npmIssuesActionProvider.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  it("provideCodeActions", () => {
+    const actions: CodeAction[] = provider.provideCodeActions(
+      textDocument,
+      range,
+      context,
+      token
+    );
+    expect(actions).to.have.lengthOf(1);
```
bd82, line 43: should there be a more detailed assertion here?

packages/vscode-dependencies-validation/test/diagnostics.spec.ts
```diff
@@ -0,0 +1,84 @@
+import proxyquire = require("proxyquire");
+import { diagnosticCollection } from "./vscodeMocks";
+import { createSandbox, SinonMock, SinonSandbox } from "sinon";
```
bd82, line 3: import order (3rd party **before** relative paths)

packages/vscode-dependencies-validation/test/diagnostics.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+import { refreshDiagnostics } from "../src/diagnostics";
+import { range } from "lodash";
+import { NPM_DEPENDENCY_ISSUES_CODE } from "../src/constants";
+
+describe("diagnostics unit test", () => {
+  const packageJsonPath = "/root/folder/package.json";
+  const Uri = {
```
bd82, line 10: on first look I think the "fuller" mocks with actual basic implementation are cleaner and better as we don't have to assume
the code does not actually need implementation details.

packages/vscode-dependencies-validation/test/diagnostics.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    it("there are 2 dependency issues", async () => {
+      problems = ["missing: json-fixer@1.6.12", "missing: lodash@0.0.1"];
+      const diagnostic = new Diagnostic(
+        new Range(0, 0, 0, 10),
+        problems.join("\n"),
+        0,
+        NPM_DEPENDENCY_ISSUES_CODE
+      );
+      //const diagnostics = [diagnostic];
+      diagnosticCollectionMock
```
bd82, line 69: instead of adding the complexity of sinon mocks here I would instead suggest
to implement the basic set/get for the "DiagnosticCollection" and assert on its **contents** after **refreshDiagnosticsProxy** is called 

If you can implement it using ECMAScript Set it may be less code than the mocks handling code here
(and every other test that would need to mock it too...).

packages/vscode-dependencies-validation/test/diagnostics.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+          Uri,
+          "@noCallThru": true,
+        },
+        "@sap-devx/npm-dependencies-validation": {
+          findDependencyIssues: async () => Promise.resolve({ problems }),
+          "@noCallThru": true,
+        },
+      });
+
+      refreshDiagnosticsProxy = diagnosticsModule.refreshDiagnostics;
```
bd82, line 45: when you init **refreshDiagnosticsProxy** only **once** in before
you make an assumption that ""../src/diagnostics"" is stateless.
This may be true now, but would it be true in the future too?
So might be safer to init it in **beforeEach**

packages/vscode-dependencies-validation/test/vscodeMocks.ts
```diff
@@ -0,0 +1,189 @@
...
+  Position,
+  Range,
+  TextDocument,
+  TextLine,
+  CodeActionKind,
+} from "vscode";
+import { VscodeOutputChannel } from "../src/vscodeTypes";
+
+export const outputChannel: VscodeOutputChannel = {
+  append(value: string) {},
```
bd82, line 18: if you add console.logs here we may get some useful tracing during tests.
or maybe it would make the test output ugly... worth a shot

packages/vscode-dependencies-validation/test/vscodeMocks.ts
```diff
@@ -0,0 +1,189 @@
...
+};
+
+export const diagnosticCollection: DiagnosticCollection = {
+  name: "",
+  // @ts-expect-error -- https://stackoverflow.com/questions/68799234/typescript-pick-only-specific-method-from-overload-to-be-passed-to-parameterst
+  set: function (
+    uri: Uri,
+    diagnostics: readonly Diagnostic[] | undefined
+  ): void {
+    //throw new Error("Function not implemented.");
```
bd82, line 31: if you can implement it naively with 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

you can make clearer assertions during tests which would also be easier to debug (debug collection object vs debug mocks/spys)

packages/vscode-dependencies-validation/test/vscodeMocks.ts
```diff
@@ -0,0 +1,189 @@
+import { NPM_DEPENDENCY_ISSUES_CODE } from "../src/constants";
```
bd82, line 1: its not a huge deal, but needing to implement these VSCode APIs makes me wonder
if we should (later) make an attempt to use VSCode-Test or https://github.com/SAP/vscode-test-cli
and get those implementations for free as the **unit** tests would be running in an env that has VSCode

packages/vscode-dependencies-validation/test/vscodeMocks.ts
```diff
@@ -0,0 +1,189 @@
...
+  },
+  has: function (uri: Uri): boolean {
+    throw new Error("Function not implemented.");
+  },
+  dispose: function (): void {
+    throw new Error("Function not implemented.");
+  },
+};
+
+export const textDocument: TextDocument = {
```
bd82, line 60: does this contain what we need for TextDocument?
https://www.npmjs.com/package/vscode-languageserver-textdocument

packages/vscode-dependencies-validation/test/vscodeMocks.ts
```diff
@@ -0,0 +1,189 @@
...
+  translate: function (lineDelta?: number, characterDelta?: number): Position {
+    throw new Error("Function not implemented.");
+  },
+  // @ts-expect-error -- https://stackoverflow.com/questions/68799234/typescript-pick-only-specific-method-from-overload-to-be-passed-to-parameterst
+  with: function (line?: number, character?: number): Position {
+    throw new Error("Function not implemented.");
+  },
+};
+
+export const range: Range = {
```
bd82, line 131: you may be able a generic range implementation that fits or mostly fits this API

packages/vscode-dependencies-validation/test/vscodeMocks.ts
```diff
@@ -0,0 +1,189 @@
...
+  union: function (other: Range): Range {
+    throw new Error("Function not implemented.");
+  },
+  // @ts-expect-error -- https://stackoverflow.com/questions/68799234/typescript-pick-only-specific-method-from-overload-to-be-passed-to-parameterst
+  with: function (start?: Position, end?: Position): Range {
+    throw new Error("Function not implemented.");
+  },
+};
+
+export const context: CodeActionContext = {
```
bd82, line 154: as opposed to VSCode like classes, this looks like static data for a specific test
should it be part of the test ?

bd82: even at cost of slight duplication?

bd82: if not, maybe at the very least freeeze it so one test won't modify it for the next one

packages/vscode-dependencies-validation/test/vscodeMocks.ts
bd82: one disadvantage of this approach is that we may end up with placeholder classes we no longer need
when the source code / tests gets refactored.

we need to discuss these two approaches online next week...

packages/vscode-dependencies-validation/test/commands.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  "@noCallThru": true,
+};
+
+describe("commands unit test", () => {
+  let sandbox: SinonSandbox;
+  let outputChannelMock: SinonMock;
+  let npmDepsValidationMock: SinonMock;
+  let diagnosticsMock: SinonMock;
+
+  before(() => {
```
bd82, line 28: what does it mean that the sandbox is shared between multiple tests?
could it cause shared state between tests?

packages/vscode-dependencies-validation/test/commands.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+      outputChannelMock
+        .expects("appendLine")
+        .withExactArgs(`\nFixing dependency issues ...`);
+      const config: NpmCommandConfig = {
+        commandArgs: ["install"],
+        cwd: dirname(packageJsonPath),
+      };
+      npmDepsValidationMock
+        .expects("invokeNPMCommand")
+        .withExactArgs(config, outputChannel)
```
bd82, line 72: i think that in-lining the config here would make the assertion easier to read

packages/vscode-dependencies-validation/test/commands.spec.ts
```diff
@@ -0,0 +1,111 @@
...
```
bd82, line 100: I think that in-lining the config here would make the assertions easier to read

packages/vscode-dependencies-validation/test/commands.spec.ts
```diff
@@ -0,0 +1,111 @@
+import { dirname } from "path";
```
bd82, line 1: I am not a big fan of sinon, but I cant argue with the result that these tests are very clear and understandable 👍

packages/npm-dependencies-validation/package.json
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/depIssues.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/types.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/types.ts
slavik-lvovsky: I was not sure I need it.
Removed.

packages/npm-dependencies-validation/src/types.ts
slavik-lvovsky: Using this syntax I removed code duplication.
Currently we do not need all these properties, but we will probably need them in the future.
Should I remove them for now?

packages/npm-dependencies-validation/src/utils/npmUtil.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/utils/npmUtil.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: I do not think we need log here, the question is what should we to return in this case ?

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: no I do not use fs-extra, I think it is redundant for this package

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: this is the recommended way for this check.
I removed the WRITE permission check for now

packages/npm-dependencies-validation/src/utils/packageJsonUtil.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/test/depIssues.spec.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/test/depIssues.spec.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
slavik-lvovsky: Done

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
slavik-lvovsky: It will not work. According to chai-as-promised API I need to use return before invokeNPMCommand

packages/npm-dependencies-validation/src/utils/npmUtil.ts
slavik-lvovsky: it is always one big chunk.
I have tested it on relatively big project and it worked

packages/npm-dependencies-validation/test/utils/packageJsonUtil/packageJsonUtil.spec.ts
slavik-lvovsky: I have added another test that checks this flow

packages/vscode-dependencies-validation/package.json
slavik-lvovsky: Done

packages/vscode-dependencies-validation/package.json
slavik-lvovsky: Removed

packages/vscode-dependencies-validation/package.json
slavik-lvovsky: removed

packages/vscode-dependencies-validation/src/commands.ts
slavik-lvovsky: Yes, I know

packages/vscode-dependencies-validation/src/commands.ts
slavik-lvovsky: I think it is not possible because we run the "npm ls --depth=0" with --json option.

packages/vscode-dependencies-validation/src/commands.ts
slavik-lvovsky: Please explain me what do you mean

packages/vscode-dependencies-validation/src/commands.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/commands.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/diagnostics.ts
slavik-lvovsky: I need this packageJsonPath as input to command. The command is constructed based on a diagnostics.
I did not find a better way to pass it

packages/vscode-dependencies-validation/src/diagnostics.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param dependencyIssueDiagnostics diagnostic collection
+ */
+export async function refreshDiagnostics(
+  packageJsonPath: string,
+  dependencyIssueDiagnostics: DiagnosticCollection
+): Promise<void> {
+  const npmLsResult = await findDependencyIssues(packageJsonPath);
+
+  const diagnostics: Diagnostic[] = [];
+  if (npmLsResult.problems) {
```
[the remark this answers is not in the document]
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/diagnostics.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/diagnostics.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/extension.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/extension.ts
slavik-lvovsky: Fixed

packages/vscode-dependencies-validation/src/extension.ts
slavik-lvovsky: I have redefined the pattern but here we still need to use this one

packages/vscode-dependencies-validation/src/extension.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/extension.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/extension.ts
slavik-lvovsky: yes

packages/vscode-dependencies-validation/test/npmIssuesActionProvider.spec.ts
slavik-lvovsky: ok

packages/vscode-dependencies-validation/test/npmIssuesActionProvider.spec.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/test/npmIssuesActionProvider.spec.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/test/diagnostics.spec.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/test/diagnostics.spec.ts
slavik-lvovsky: I changed it, but it case it will has state the tests will probably fail

packages/vscode-dependencies-validation/test/diagnostics.spec.ts
slavik-lvovsky: Not sure that your approach is better for a developer using sinon. Let's talk about it

packages/vscode-dependencies-validation/test/commands.spec.ts
slavik-lvovsky: it is ok, see here: https://sinonjs.org/releases/latest/sandbox/

packages/vscode-dependencies-validation/test/commands.spec.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/test/commands.spec.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/test/commands.spec.ts
slavik-lvovsky: Agree

packages/vscode-dependencies-validation/test/vscodeMocks.ts
slavik-lvovsky: Not sure, let's talk about it

packages/vscode-dependencies-validation/src/extension.ts
slavik-lvovsky: Done

PR #155, opened 2021-12-07
Auto fix versions misalignment

slavik-lvovsky:


examples/scheduled-actions-workspace/vscode-settings-actions2/package.json
```diff
@@ -9,9 +9,10 @@
   },
   "main": "index.js",
   "dependencies": {
-    "json-fixer": "1.6.13"
+    "json-fixer": "1.6.12",
+    "lodash": "4.17.21"
   },
   "devDependencies": {
-    "lodash": "4.17.21"
+    "lodash": "^4.0.21"
```
bd82, line 16: are the multiple versions of lodash supposed to be here?

nyc.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   "check-coverage": true,
   all: true,
   include: "**/src/**",
+  exclude: "**/src/extension.ts",
```
bd82, line 6: I think this exclusion does not belong in the **root** config
rather in the specific sub-package config as not all sub-packages have an extension.ts

packages/npm-dependencies-validation/src/utils/npmUtil.ts
```diff
@@ -27,9 +30,42 @@ export function invokeNPMCommand<T>(
...
+    // TODO: does not stop here when an error occurs ?
+    command.on("error", (error) => {
+      const { stack } = error;
+      sendDataToOutputChannel(`${stack}`, outputChannel);
+      reject(error);
+    });
+  });
+}
+
+function executeSpawn(
```
bd82, line 58: `executeSpawn` is a bad name because there is a similar API
to `spawn` in its own module (child-process) called `exec`

maybe `invokekSpawn`?

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
```diff
@@ -13,15 +17,35 @@ describe("npmUtil unit test", () => {
...
+      const appendSpy: SinonSpy = sandbox.spy(outputChannel, "append");
       const config = { commandArgs: ["ls"], cwd: "./non_existing_path" };
-      return expect(invokeNPMCommand(config)).to.be.rejectedWith(
-        `spawn ${getNPM()} ENOENT`
+      return invokeNPMCommandWithJsonResult(config, outputChannel)
+        .then(() => {
+          assert.fail("test should fail");
+        })
+        .catch((error: Error) => {
+          expect(appendSpy.called).to.be.true;
```
bd82, line 29: should we assert on the messages passed to `outputChannel`?

packages/npm-dependencies-validation/test/utils/npmUtil.spec.ts
```diff
@@ -13,15 +17,35 @@ describe("npmUtil unit test", () => {
...
+    });
+
+    it("invokeNPMCommandWithJsonResult succeeded with ls", async () => {
+      const appendSpy: SinonSpy = sandbox.spy(outputChannel, "append");
+      const config = { commandArgs: ["ls", "--depth=0"], cwd: "./" };
+      const result: NpmLsResult = await invokeNPMCommandWithJsonResult(
+        config,
+        outputChannel
       );
+      expect(appendSpy.called).to.be.true;
```
bd82, line 41: should we assert on the messages passed to outputChannel?

packages/vscode-dependencies-validation/package.json
```diff
@@ -18,10 +18,31 @@
...
+          "description": "Delays automatic dependencies misalignment fixing on startup (seconds)"
+        }
+      }
+    }
+  },
   "scripts": {
     "ci": "npm-run-all clean compile bundle package",
     "clean": "rimraf ./dist *.vsix ./coverage ./.nyc_output",
     "compile": "tsc -p ./",
+    "watch": "tsc -p ./ --watch",
```
bd82, line 45: I asked somewhere else, but maybe have not noticed the reply there.
do we need a compile watch script here?
considering we have a root level watch for the whole mono-repo?

packages/vscode-dependencies-validation/src/autofix/configuration.ts
```diff
@@ -0,0 +1,18 @@
+import { VscodeWorkspace } from "../vscodeTypes";
+
+export const ENABLE_AUTOFIX = "dependenciesValidation.enableAutoFix";
+const DELAY_AUTOFIX = "dependenciesValidation.delayAutoFix";
+
+export function isAutoFixEnabled(workspace: VscodeWorkspace): boolean {
```
bd82, line 6: I think a **defaults** object which would allow only a single getConfig method.
could make this code cleaner (assuming there would be more user settings in the future).

I will try to implement this, there could be issues with the type information though...

bd82: WDYT?

```typescript
import { workspace } from "vscode"; // change this to avoid direct dep on vscode
type VscodeWorkspace = typeof workspace

export const ENABLE_AUTOFIX = "dependenciesValidation.enableAutoFix";
export const DELAY_AUTOFIX = "dependenciesValidation.delayAutoFix";

interface ExtensionConfiguration {
    [ENABLE_AUTOFIX] : boolean
    [DELAY_AUTOFIX] : number
}

type ExtConfigKey = keyof  ExtensionConfiguration

const ExtensionConfigDefaults: ExtensionConfiguration = {
    [ENABLE_AUTOFIX] : false,
    [DELAY_AUTOFIX] : 0
}

export function getConfigValue<T extends ExtConfigKey>(key: T, workspace:VscodeWorkspace): ExtensionConfiguration[T] {
    const workspaceConfig = workspace.getConfiguration();
    const value =  workspaceConfig.get(key, ExtensionConfigDefaults[key])
    return value;
}


```

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
```diff
@@ -0,0 +1,59 @@
...
+  getAutoFixDelay,
+  isAutoFixEnabled,
+} from "./configuration";
+import { addPackageJsonFileWatcher as addPackageJsonFileWatcher } from "./packageJsonFileWatcher";
+import { addUnsupportedFilesWatcher } from "./unsupportedFilesWatcher";
+import { findAndFixDepsIssues } from "../util";
+import { clearDiagnostics } from "../util";
+
+export function activateDepsIssuesAutoFix(
+  vscodeConfig: VscodeFileEventConfig & VscodeUriFile
```
bd82, line 18: are you  using the DI in method params because you find it more convinient?
- e.g: for tests?

I remember you mentioned `proxyQuire` replaced must of the need for this DI...

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
```diff
@@ -0,0 +1,59 @@
...
+  addPackageJsonFileWatcher(vscodeConfig);
+  addUnsupportedFilesWatcher(vscodeConfig);
+
+  onAutoFixChange(vscodeConfig);
+}
+
+function fixWorkspaceDepsIssues(vscodeConfig: VscodeFileEventConfig): void {
+  setTimeout(() => {
+    void findDepsIssuesFixThemAndCleanProblemsView(vscodeConfig);
+  }, getAutoFixDelay(vscodeConfig.workspace));
```
bd82, line 30: you need to multiply the autoFix delay by 1,000 to transform to seconds

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+async function findDepsIssuesFixThemAndCleanProblemsView(
+  vscodeConfig: VscodeFileEventConfig
+): Promise<void> {
+  const { workspace, diagnosticCollection, outputChannel } = vscodeConfig;
+  if (!isAutoFixEnabled(workspace)) return;
+
+  const packageJsonUris = await getPackageJsonUris(workspace);
+  // TODO: should we do it in parrallel or sequentially ???
```
bd82, line 53: if we do it in parallel the output channel output would not be legible

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
```diff
@@ -0,0 +1,59 @@
+import type { Uri } from "vscode";
```
bd82, line 1: we might want to consider a different name for this file
as we already have `depIssues` (even though it is in another package)

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
```diff
@@ -0,0 +1,59 @@
...
+): Promise<void> {
+  const { workspace, diagnosticCollection, outputChannel } = vscodeConfig;
+  if (!isAutoFixEnabled(workspace)) return;
+
+  const packageJsonUris = await getPackageJsonUris(workspace);
+  // TODO: should we do it in parrallel or sequentially ???
+  // currently it works sequentially
+  for (const uri of packageJsonUris) {
+    await findAndFixDepsIssues(uri, outputChannel);
+    clearDiagnostics(diagnosticCollection, uri);
```
bd82, line 57: in regards to `clearDiagnostics` 
Why are diagnostics even involved in `auto-fix` flow?
don't they only show for opened files? and autofix works on all files in workspace?

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
```diff
@@ -0,0 +1,59 @@
...
+} from "./configuration";
+import { addPackageJsonFileWatcher as addPackageJsonFileWatcher } from "./packageJsonFileWatcher";
+import { addUnsupportedFilesWatcher } from "./unsupportedFilesWatcher";
+import { findAndFixDepsIssues } from "../util";
+import { clearDiagnostics } from "../util";
+
+export function activateDepsIssuesAutoFix(
+  vscodeConfig: VscodeFileEventConfig & VscodeUriFile
+): void {
+  fixWorkspaceDepsIssues(vscodeConfig);
```
bd82, line 20: There seems to be a logical problem here.
Autofix would only run on Ext  startup
and when the auto-fix boolean config changes.

But should not it not run **anytime** there is a mis-alignemnt / missing dep?

bd82: e.g due to `rm -rf node_modules`

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
```diff
@@ -0,0 +1,59 @@
...
+  }, getAutoFixDelay(vscodeConfig.workspace));
+}
+
+function getPackageJsonUris(workspace: VscodeWorkspace): Thenable<Uri[]> {
+  return workspace.findFiles("package.json", "**​/node_modules/**");
+}
+
+function onAutoFixChange(vscodeConfig: VscodeFileEventConfig): void {
+  vscodeConfig.workspace.onDidChangeConfiguration((event) => {
+    const affected = event.affectsConfiguration(ENABLE_AUTOFIX);
```
bd82, line 39: logically a configuration change may only require handling if the value was changed to
`true`? 

I think the above statement may be more correct if we remove any problems view handling from this flow.

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
bd82: BTW when you implement this I beleive that a trigger that causes autofix flow
to start should be delayed for several minutes (e.g 5-10) because otherwise
flows such as "rm -rf node_modules" or "yo my-generator" will cause race conditionons
with the auto-fix tool.

bd82: So I see now some of the trigger event flows are handled in `addPackageJsonFileWatcher`

It was a little unclear to me, but I am not yet sure if I am the problem or if the code organization can be imporved and if so how...

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
```diff
@@ -0,0 +1,59 @@
...
+import { addUnsupportedFilesWatcher } from "./unsupportedFilesWatcher";
+import { findAndFixDepsIssues } from "../util";
+import { clearDiagnostics } from "../util";
+
+export function activateDepsIssuesAutoFix(
+  vscodeConfig: VscodeFileEventConfig & VscodeUriFile
+): void {
+  fixWorkspaceDepsIssues(vscodeConfig);
+  addPackageJsonFileWatcher(vscodeConfig);
+  addUnsupportedFilesWatcher(vscodeConfig);
```
bd82, line 22: My understanding is that `addUnsupportedFilesWatcher` is meant to handle the edge case
of a package.json file which stopped being **exlcuded** due to the existance of none npm files (e.g yarn.lock).

My Feedback:
- Long term watcher logic will be insufficent because we may have more complex conditions
  - on when/what is being excluded.
- evaluate removing this "unsupported' watcher logic completely and **replacing** it with periodic checks (e.g hourly). reasons:
   - Reduce solution complexity.
   - Would still be correct, just delayed.
   - Avoids future problem of expanding advanced detection of unsupported flows.
   - Periodic checks will also detect edge cases we have not considered.
   
One problem with periodic flows is that it could cause race conditions with user operations.
- e.g a periodic check happening just when an end user wrote `rm -rf node_modules` but before they did `npm install` again.

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+function fixWorkspaceDepsIssues(vscodeConfig: VscodeFileEventConfig): void {
+  setTimeout(() => {
+    void findDepsIssuesFixThemAndCleanProblemsView(vscodeConfig);
+  }, getAutoFixDelay(vscodeConfig.workspace));
+}
+
+function getPackageJsonUris(workspace: VscodeWorkspace): Thenable<Uri[]> {
+  return workspace.findFiles("package.json", "**​/node_modules/**");
```
bd82, line 34: random thought:
is there a value in filtering?
- edge case: a package.json file which does not meet the schema.
- edge case: a package.json which has no deps and dev-deps --> can be ignored.
- optimization: knowing for each package.json if it has deps/devDeps/both for our `npm ls` flows.

packages/vscode-dependencies-validation/test/moduleProxies.ts
```diff
@@ -12,5 +12,38 @@ export const npmDepsValidationProxy = {
...
+};
+
+export const configurationProxy = {
+  isAutoFixEnabled() {
+    throw new Error("configurationProxy method is not implemented");
+  },
+  "@noCallThru": true,
+};
+
+export const utilProxy = {
```
bd82, line 35: so it seems strange that we need to proxy our own module, particularly when it
is not using un-aviliable imports (`vscode`). So I looked at one of the tests that uses 
this proxy and it does not look like a "Black Box" unit test, so maybe this proxy won't be needed.

I will give more details on a specific test in another comment

packages/vscode-dependencies-validation/test/autofix/eventUtilspec.ts
```diff
@@ -0,0 +1,69 @@
...
+  context("handlePackageJsonEvent()", () => {
+    it("autofix is disabled", async () => {
+      const uri = <Uri>{ fsPath: "root/folder/project/package.json" };
+      configurationProxySinonMock.expects("isAutoFixEnabled").returns(false);
+      utilProxySinonMock.expects("findAndFixDepsIssues").never();
+
+      await handlePackageJsonEventProxy(uri, vscodeConfig);
+    });
+
+    it("path is in node_modules", async () => {
```
bd82, line 42: I don't think this is a **maintainable** test.

- It seems to test an **flow** instead of being "unity".
- It utilizes very specific knowlage (whitebox) of which methods would be called **trasitivily** due to the invoking  `handlePackageJsonEventProxy`
- It "translates" this **whitebox** information on the proxied methods expected return values 
  to test assertions that "prove" correct behaivor.
- and it refers to these methods via **string** names, so future refactoring would break the test.

Maybe the events flows are not easily testable without more E2E / Integration testing capabilities.
We need to investigate this, but I suspect this style of test will cause high TCO over time.

packages/vscode-dependencies-validation/test/autofix/eventUtilspec.ts
```diff
@@ -0,0 +1,69 @@
+import type { Uri } from "vscode";
+import * as proxyquire from "proxyquire";
+import { VscodeFileEventConfig } from "../../src/vscodeTypes";
+import { handlePackageJsonEvent } from "../../src/autofix/eventUtil";
+import { configurationProxy, utilProxy } from "../moduleProxies";
+import { SinonMock, createSandbox, SinonSandbox } from "sinon";
+
+describe("eventUtil unit tests", () => {
```
bd82, line 8: See the comment below, I think we need to investigate it in more depth
if it is possible to easily build & **maintain** unit tests for such complex event related flows

packages/vscode-dependencies-validation/test/diagnostics.spec.ts
```diff
@@ -56,24 +53,24 @@ describe("diagnostics unit test", () => {
     });
 
     it("there are 2 dependency issues", async () => {
+      const uri = <Uri>{ fsPath: packageJsonPath };
```
bd82, line 56: these tests are understandable but could be better if the code they are testing was refactored.

```typescript
export async function refreshDiagnostics(
  uri: Uri,
  dependencyIssueDiagnostics: DiagnosticCollection
): Promise<void> {
  if (isNotInNodeModules(uri)) {
    const { problems } = await findDependencyIssues(uri.fsPath);

    const diagnostics = isEmpty(problems)
      ? []
      : [constructDiagnostic(problems)];

    dependencyIssueDiagnostics.set(uri, diagnostics);
  }
}
```

1. the filtering of `isNotInNodeModules` could be extracted outside this function
  - generally could we ever trigger a check for anything that is **inside** node_modules? is this filter really needed?
  - but lets assume it is needed
2. `findDependencyIssues` could be defined as a DI parameter so the deps of this function are more clear.
3. `dependencyIssueDiagnostics` could be defined as a **sub-set** of `DiagnosticCollection` APIs

Then we will have something like:

```typescript
// TODO; consider config object instead of three params
export async function refreshDiagnostics(
  uri: Uri,
  dependencyIssueDiagnostics: Pick<DiagnosticCollection, "set">,
  findDependencyIssues: async (path:string) =>  {problems:string[] }
): Promise<void> {
    // more explicit early exit.
    if (isInsideNodeModules(uri)) { 
        return;
    }

    const { problems } = await findDependencyIssues(uri.fsPath);
    const diagnostics = isEmpty(problems)
      ? []
      : [constructDiagnostic(problems)];
    dependencyIssueDiagnostics.set(uri, diagnostics);
}
```

Now this is easier to test without needing **proxyQuire** for anything but `vscode`.
The problem is that **proxyQuire** may need to be updated for any new import at the root of module.
- e.g if that new import transitivily needs `vscode` 
- or if the **unit** we are testing is using that new import.

And we won't automagically know if a test need to be updated due to production code changes (adding new import at the module level). But if a **unit** (function) needs another "DI" param, any test that calls it directly would fail **during TypeScript compilation**.

The risk of course is that we would need to pass too many of these DI params between functions
- but that is not the case here.
- This raises the question of evaluating using  proper DI framework for writing our VSCode exts (out of scope...)

Now you could write a simpler test:

```typescript

let refreshDiagnosticsProxy: any // todo: missing right type

beforeEach(() => {
      const diagnosticsModule = proxyquire("../src/diagnostics", {
        // Note: only proxying vscode
        vscode: {
          Range,
          Diagnostic,
          "@noCallThru": true,
        },
      });
      refreshDiagnosticsProxy = diagnosticsModule.refreshDiagnostics;
})

it("releant path outside node_modules", async () => {
      // note: preparations of the params are fully isolated in this code block
     //           without any external helpers / proxies.
      const uri = <Uri>{ fsPath: "/user/home/projects/app-studio-toolkit };
      const issues = ["missing: json-fixer@1.6.12", "missing: sometingelse@1.2.3];
      const findDependencyIssues: async () => isssues
      const diagnosticCollection = new Set<Uri, string[]>()

      // note the `Set` should match the DiagonsticColletion set method
      await refreshDiagnosticsProxy(uri, diagnosticCollectionMock, findDependencyIssues);

      // Note: we can assert on the Set object as it was modified, no need for Sinon here...
      const diagnosticsEntryResult = diagnosticCollection.get(uri);
      expect(diagnosticsEntryResult).to.have.lengthOf(2);
      expect(diagnosticsEntryResult).to.deep.eql(issues)
});

// now we test **inside** node modules
it("releant path inside node_modules", async () => {
      const uri = <Uri>{ fsPath: "/user/home/projects/app-studio-toolkit };
      const findDependencyIssues: async () => []
      const diagnosticCollection = new Set<Uri, string[]>()

      await refreshDiagnosticsProxy(uri, diagnosticCollectionMock, findDependencyIssues);
      expect(diagnosticCollection.has(uri)).to.be.false;
});
```

WDYT?

packages/vscode-dependencies-validation/src/autofix/eventUtil.ts
```diff
@@ -0,0 +1,23 @@
...
+import { VscodeFileEventConfig, VscodeWorkspace } from "../vscodeTypes";
+
+export async function handlePackageJsonEvent(
+  uri: Uri,
+  vscodeConfig: VscodeFileEventConfig
+): Promise<void> {
+  const { workspace, diagnosticCollection, outputChannel } = vscodeConfig;
+  if (canBeFixed(workspace, uri)) {
+    await findAndFixDepsIssues(uri, outputChannel);
+    clearDiagnostics(diagnosticCollection, uri);
```
bd82, line 17: I think there is an hidden assumption here that `findAndFixDepsIssues` always succeeded and that the 
`clearDiagnostics` is always safe/correct to execute.

packages/vscode-dependencies-validation/src/autofix/packageJsonFileWatcher.ts
```diff
@@ -0,0 +1,30 @@
...
+import { handlePackageJsonEvent } from "./eventUtil";
+
+const debouncedHandlePackageJsonEvent = debounce(handlePackageJsonEvent, 3000);
+
+// TODO: what should happen after git clone ??
+export function addPackageJsonFileWatcher(
+  vscodeConfig: VscodeFileEventConfig
+): void {
+  const fileWatcher =
+    vscodeConfig.workspace.createFileSystemWatcher("**/package.json");
```
bd82, line 13: the glob pattern seems to have some negate options.
perhaps that can be used to ignore package.json files inside node_modules
without needing a post-processing filter?
 - https://code.visualstudio.com/api/references/vscode-api#GlobPattern

packages/vscode-dependencies-validation/src/autofix/packageJsonFileWatcher.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  fileWatcher.onDidChange(onChange(vscodeConfig));
+  fileWatcher.onDidCreate(onCreate(vscodeConfig));
+}
+
+function onCreate(vscodeConfig: VscodeFileEventConfig): any {
+  return (uri: Uri) => handlePackageJsonEvent(uri, vscodeConfig);
```
bd82, line 20: I think that if a file is created and immediately changed you may end up running
two heavy operations in parallel, it may be safer to always debounce with `debouncedHandlePackageJsonEvent`

packages/vscode-dependencies-validation/src/autofix/unsupportedFilesWatcher.ts
```diff
@@ -0,0 +1,50 @@
+import type { Uri } from "vscode";
```
bd82, line 1: as mentioned elsewhere in this review, I think we should consider keeping this scenario out-of-scope.
as the edge case which consists of:
- remove exclusion (rm yarn.lock)
- having errors to show.

Seems somewhat rare.

Also:
- someone may just be re-generating their yarn.lock so running npm -i in that case would be a mistake.
- if we implement some periodic (hourly) checks it will still handle the edge case of changing from yarn -> npm

packages/vscode-dependencies-validation/src/commands.ts
```diff
@@ -1,38 +1,40 @@
...
   outputChannel: VscodeOutputChannel,
-  packageJsonPath: string,
-  dependencyIssuesDiagnosticCollection: DiagnosticCollection
+  diagnosticCollection: DiagnosticCollection,
+  uri: Uri
 ): Promise<void> {
+  // switched to outputchannel only in manual mode
   outputChannel.show(true);
+  await fixDepsIssues(uri, outputChannel);
+  clearDiagnostics(diagnosticCollection, uri);
```
bd82, line 14: Again there is an assumption here that the `fixDepsIssues` succeeded

packages/vscode-dependencies-validation/src/diagnostics.ts
```diff
@@ -1,44 +1,39 @@
...
+import { isNotInNodeModules } from "./util";
 
 /**
- * Analyzes the package.json text document for problems.
- * @param doc package.json text document to analyze
+ * Analyzes package.json file for problems.
+ * @param packageJsonPath package.json file path to analyze
  * @param dependencyIssueDiagnostics diagnostic collection
  */
 export async function refreshDiagnostics(
```
bd82, line 13: I commented about this method in one of its tests

packages/vscode-dependencies-validation/src/editorChanges.ts
```diff
@@ -0,0 +1,63 @@
...
+import { clearDiagnostics } from "./util";
+import { VscodePackageJsonChangesConfig } from "./vscodeTypes";
+
+export function subscribeToPackageJsonChanges(
+  vscodeConfig: VscodePackageJsonChangesConfig
+): void {
+  const { window, subscriptions, workspace, diagnosticCollection } =
+    vscodeConfig;
+
+  if (window.activeTextEditor) {
```
bd82, line 17: can you link to the template you used for this method?

bd82: I mean in a comment

packages/vscode-dependencies-validation/src/util.ts
```diff
@@ -0,0 +1,46 @@
...
+import { VscodeOutputChannel } from "./vscodeTypes";
+
+const NOT_IN_NODE_MODULES_PATTERN =
+  /^(?!.*[\\|\/]node_modules[\\|\/]).*[\\|\/].+/;
+
+export function isNotInNodeModules(uri: Uri): boolean {
+  return NOT_IN_NODE_MODULES_PATTERN.test(uri.fsPath);
+}
+
+export function clearDiagnostics(
```
bd82, line 17: What value does this one liner function provide?

packages/vscode-dependencies-validation/src/util.ts
```diff
@@ -0,0 +1,46 @@
...
+import { dirname } from "path";
+import {
+  findDependencyIssues,
+  invokeNPMCommand,
+} from "@sap-devx/npm-dependencies-validation";
+import { isEmpty } from "lodash";
+import { VscodeOutputChannel } from "./vscodeTypes";
+
+const NOT_IN_NODE_MODULES_PATTERN =
+  /^(?!.*[\\|\/]node_modules[\\|\/]).*[\\|\/].+/;
```
bd82, line 11: this pattern may be too complex

1. it includes the parts around the `/node_modules/` (before and after)
2. it includes both types of file separators, but you can build the pattern dynamically with path.sep
   - https://nodejs.org/api/path.html#pathsep
 3. it uses negative lookahead assertion, but that a complex regexp feature, just test for `/node_modules/` in the **base** path and and flip the condition....

packages/vscode-dependencies-validation/src/util.ts
```diff
@@ -0,0 +1,46 @@
...
+  return fixDepsIssues(packageJsonUri, outputChannel);
+}
+
+export async function fixDepsIssues(
+  packageJsonUri: Uri,
+  outputChannel: VscodeOutputChannel
+): Promise<void> {
+  // package.json parent dir
+  const cwd = dirname(packageJsonUri.fsPath);
+  outputChannel.appendLine(`\nFixing dependency issues of ${cwd} ...\n`);
```
bd82, line 40: instead of CWD print the full path to pkg.json ?

packages/vscode-dependencies-validation/src/util.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const config = { commandArgs: ["install"], cwd };
+  await invokeNPMCommand(config, outputChannel);
+
+  outputChannel.appendLine(`Done.\n`);
```
bd82, line 45: add the full path to the done message so it would be easy to match start->end sections

packages/vscode-dependencies-validation/src/vscodeTypes.ts
```diff
@@ -0,0 +1,62 @@
...
+  CodeActionKind,
+  DiagnosticCollection,
+} from "vscode";
+
+export type VscodeOutputChannel = Pick<
+  OutputChannel,
+  "append" | "show" | "appendLine"
+>;
+
+export type VscodeWorkspace = typeof workspace;
```
bd82, line 18: using such "wide" types that include everything in some VSCode property
can result in more complex tests / mocks

packages/vscode-dependencies-validation/test/autofix/configuration.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      const res = isAutoFixEnabled(workspaceMock);
+      expect(res).to.be.false;
+    });
+
+    it("autofix is enabled", () => {
+      workspaceConfigurationSinonMock
+        .expects("get")
+        .withExactArgs(ENABLE_AUTOFIX, false)
+        .returns(true);
+      const res = isAutoFixEnabled(workspaceMock);
```
bd82, line 46: why do we need sinon and assertions on how isAutoFixEnabled implemented to test this?
Can't we just provide a matching placeholder implementation directly to isAutoFixEnabled
and only assert on the final output?

packages/vscode-dependencies-validation/test/autofix/configuration.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      expect(res).to.be.true;
+    });
+  });
+
+  context("getAutoFixDelay()", () => {
+    it("delay is 30 seconds", () => {
+      const delay = 30;
+      workspaceConfigurationSinonMock
```
bd82, line 54: same question as the above comment

Note I also suggested a another way to implement the configurations 
so maybe the tests would change in this case

packages/vscode-dependencies-validation/test/commands.spec.ts
```diff
@@ -1,101 +1,58 @@
...
     outputChannelSinonMock.verify();
-    npmDepsValidationMock.verify();
+    utilProxySinonMock.verify();
+    sandbox.restore();
   });
 
-  context("executeAllFixCommand()", () => {
-    let fixAllDepIssuesCommandProxy: typeof fixAllDepIssuesCommand;
+  context("fixProjectDepsIssues()", () => {
+    let fixProjectDepsIssuesProxy: typeof internal.fixProjectDepsIssues;
```
bd82, line 26: cool usage of typescript 👍

packages/vscode-dependencies-validation/test/autofix/unsupportedFilesWatcher.spec.ts
```diff
@@ -0,0 +1,120 @@
+import type {
```
bd82, line 1: I am not diving into this test in-depth right now.
I think this has the same questions of blackbox vs whitebox unit tests + mocks
and what are the alternatives when we are trying to test "events" / "mini flows"

packages/vscode-dependencies-validation/test/autofix/packageJsonFileWatcher.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+      onChangeProxy = packageJsonFileWatcherModule.internal.onChange;
+
+      eventUtilProxySinonMock = sandbox.mock(eventUtilProxy);
+    });
+
+    after(() => {
+      eventUtilProxySinonMock.verify();
+    });
+
+    it.skip("debouncedHandlePackageJsonEvent is called", async () => {
```
bd82, line 127: skip?

packages/vscode-dependencies-validation/test/autofix/packageJsonFileWatcher.spec.ts
```diff
@@ -0,0 +1,137 @@
+import type {
```
bd82, line 1: same discussion on how to test flows/events...

bd82:
The reasons I'm being strict / critial in regards to the tests structure are because:
1. We are learning how to write better and more maintainable tests for VSCode extensions, so this naturally requires several iteration.
2. I beleive there are some solutions we could integrate in the future which would allow for easier integration tests in VSCode.
    This causes me to potentially prefer **not** writing **unit** tests for such flows if we are unable to find a good enough solution with (only unit tests ) that will **not** have a high long term TCO

examples/scheduled-actions-workspace/vscode-settings-actions2/package.json
slavik-lvovsky: I have reverted this change

nyc.config.js
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
slavik-lvovsky: ok

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
slavik-lvovsky: Not exactly. 
If a relevant package.json file is opened and has issues and auto fix runs and fixes the file, issues will still be shown in Problems view.
We need to clear them.

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
slavik-lvovsky: Let's talk about it. I do not understand what is wrong here

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
slavik-lvovsky: it will work only if auto fix is true, see findDepsIssuesFixThemAndCleanProblemsView method

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
slavik-lvovsky: I do not think we need periodic checks here. 
Let's talk about it

packages/vscode-dependencies-validation/src/autofix/depsIssues.ts
slavik-lvovsky: I did not understand what you tried to improve here.
Let's talk about it.

packages/vscode-dependencies-validation/test/autofix/eventUtilspec.ts
slavik-lvovsky: Let's talk about it.
In my opinion this is the only "unit test" that we can implement here

packages/vscode-dependencies-validation/test/diagnostics.spec.ts
slavik-lvovsky: I understand what you tried to achieve but cannot agree with this approach.
In this code you replace actual diagnosticCollection by new Set<Uri, string[]>().
Why do you believe that real DiagnosticCollection is based on Set ?
It is not more a unit test of DiagnosticCollection, it tests only Set

packages/vscode-dependencies-validation/src/autofix/eventUtil.ts
slavik-lvovsky: No, it will not reach clearDiagnostics if findAndFixDepsIssues fails

packages/vscode-dependencies-validation/src/autofix/packageJsonFileWatcher.ts
slavik-lvovsky: it is not supported. https://github.com/microsoft/vscode/issues/27051

packages/vscode-dependencies-validation/src/autofix/packageJsonFileWatcher.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/autofix/unsupportedFilesWatcher.ts
slavik-lvovsky: Let's talk about it

packages/vscode-dependencies-validation/src/commands.ts
slavik-lvovsky: If it fails it does not reach clearDiagnostics

packages/vscode-dependencies-validation/src/editorChanges.ts
slavik-lvovsky: https://github.com/microsoft/vscode-extension-samples/blob/e958fb7d45abda8177f413062318447ce59c780f/code-actions-sample/src/diagnostics.ts#L47

packages/vscode-dependencies-validation/src/util.ts
slavik-lvovsky: I use it several times in the code

packages/vscode-dependencies-validation/src/util.ts
slavik-lvovsky: path.sep does not work in regexp
Let's try it together

packages/vscode-dependencies-validation/src/util.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/util.ts
slavik-lvovsky: Done

packages/vscode-dependencies-validation/src/vscodeTypes.ts
slavik-lvovsky: I use at least 6-7 methods on this type and there are no problems with tests

packages/vscode-dependencies-validation/test/autofix/packageJsonFileWatcher.spec.ts
slavik-lvovsky: was not able to test debounced method

packages/vscode-dependencies-validation/test/autofix/configuration.spec.ts
slavik-lvovsky: I do not think it helps us. Let's talk about it

examples/scheduled-actions-workspace/vscode-settings-actions2/package.json
```diff
@@ -7,11 +7,5 @@
...
     "url": "https://github.com/SAP/app-studio-toolkit"
   },
-  "main": "index.js",
-  "dependencies": {
-    "json-fixer": "1.6.13"
-  },
-  "devDependencies": {
-    "lodash": "4.17.21"
-  }
+  "main": "index.js"
```
bd82, line 10: - sample name none descriptive?
- sample really does not need depedenceis ?

bd82: oops this is not a sample of tests, but a demo for consumers

packages/npm-dependencies-validation/src/api.ts
```diff
@@ -3,3 +3,9 @@ export { findDependencyIssues } from "./depIssues";
 export { invokeNPMCommand } from "./utils/npmUtil";
 
 export { NpmLsResult, NpmCommandConfig } from "./types";
+
+export {
+  yarnManagerFiles,
```
bd82, line 8: Note to self: should we handle the edge case of files causing exclusions?

bd82:
I am merging this and we will continue revisions on this change after merge.

PR #158, opened 2021-12-22
build: resolve failure on nodejs 12.x / 16.x

bd82:


bd82:
merging this into the feature branch because adding the missing test case (for coverage) is not enough
because production code change is needed to fix a bug as well.

PR #161, opened 2021-12-28
feat: debounce diagnostics on user input

bd82:


packages/vscode-dependencies-validation/src/autofix/eventUtil.ts
```diff
@@ -10,6 +10,9 @@ import { VscodeFileEventConfig, VscodeWorkspace } from "../vscodeTypes";
 
 export const debouncedHandleProjectChange = debounce(handleProjectChange, 3000);
 
+// TODO: would this get invoked while package.json is being edited flow?
+//   may cause duplicate "npm i"...
```
slavik-lvovsky, line 14: the cause of duplicate npm install message is not here i think

packages/vscode-dependencies-validation/src/commands.ts
```diff
@@ -8,9 +8,10 @@ async function fixProjectDepsIssues(
   diagnosticCollection: DiagnosticCollection,
   uri: Uri
 ): Promise<void> {
-  // switched to outputchannel only in manual mode
+  // switched to output-channel only in manual mode
   outputChannel.show(true);
   await fixDepsIssues(uri, outputChannel);
+  // TODO: assumes the fixing was successful, instead we should consider re-running the validation...
```
slavik-lvovsky, line 14: npm install fails in case of error

packages/vscode-dependencies-validation/src/commands.ts
```diff
@@ -26,6 +27,7 @@ export function registerCommands(vscodeConfig: VscodeCommandsConfig): void {
   );
 }
 
+// TODO: consider in-lining this function
```
slavik-lvovsky, line 30: it will be harder to test if you make it inline

packages/vscode-dependencies-validation/src/editorChanges.ts
```diff
@@ -17,12 +18,13 @@ export function subscribeToPackageJsonChanges(
 ): void {
   const { window, subscriptions, workspace, diagnosticCollection } =
     vscodeConfig;
-
-  if (window.activeTextEditor) {
-    void refreshDiagnostics(
-      window.activeTextEditor.document.uri,
-      diagnosticCollection
-    );
+  if (
```
slavik-lvovsky, line 21: Would it be better to get uri before the if and then just use it twice in the method ?

const uri  = window.activeTextEditor?.document.uri

packages/vscode-dependencies-validation/src/editorChanges.ts
```diff
@@ -48,8 +50,12 @@ function executeRefreshDiagnosticsOnEditorChange(
   diagnosticCollection: DiagnosticCollection
 ) {
   return (editor: TextEditor | undefined) => {
-    if (editor) {
-      void refreshDiagnostics(editor.document.uri, diagnosticCollection);
+    if (editor && shouldBeChecked(editor.document.uri.path)) {
+      {
+        const uri = editor.document.uri;
```
slavik-lvovsky, line 55: same as above

packages/vscode-dependencies-validation/src/editorChanges.ts
```diff
@@ -58,10 +64,15 @@ function executeRefreshDiagnosticsOnDocumentChangeEvent(
...
-    // but only once for "true" text changes
-    if (!isEmpty(event.contentChanges)) {
-      void refreshDiagnostics(event.document.uri, diagnosticCollection);
+    if (
+      // this event gets called multiple (four) times for each text change
+      // but only once for "true" text changes
+      !isEmpty(event.contentChanges) &&
+      shouldBeChecked(event.document.uri.path)
+    ) {
+      const uri = event.document.uri;
```
slavik-lvovsky, line 73: same as above

packages/vscode-dependencies-validation/src/diagnostics/debounce.ts
```diff
@@ -0,0 +1,68 @@
...
+
+// we are using a fairly long `wait` because the operation of detecting package.json
+// dependencies issues is very heavy.
+// https://ux.stackexchange.com/questions/95336/how-long-should-the-debounce-timeout-be
+const DEBOUNCE_WAIT = 2000;
+
+export function getOptimizedRefreshDiagnostics(
+  uri: Uri
+): RefreshDiagnosticsFunc {
+  if (!OPTIMIZED_PATHS_TO_FUNC.has(uri.path)) {
```
slavik-lvovsky, line 20: const { path } = uri;
path.uri is used thrice

packages/vscode-dependencies-validation/src/diagnostics/debounce.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * A naive cleanup which removes references to the debounced functions
+ */
+function garbageCollect(
+  getWorkspaceFolder: typeof workspace.getWorkspaceFolder
+): void {
+  const entries = OPTIMIZED_PATHS_TO_FUNC.entries();
+  for (const [path, { uri }] of entries) {
+    const isOutSideWorkspace = getWorkspaceFolder(uri) === undefined;
```
slavik-lvovsky, line 58: an uri come from workspace package.json, so how is it possible that an uri is not part of a workspace ?

packages/vscode-dependencies-validation/src/autofix/eventUtil.ts
bd82: Yeah, but I still want to debug this flow, it may be a bug/edge case

packages/vscode-dependencies-validation/src/commands.ts
bd82: I will add this as a comment

packages/vscode-dependencies-validation/src/diagnostics/debounce.ts
bd82: multiple roots project where some sub-roots have been added and then removed

packages/vscode-dependencies-validation/src/diagnostics/debounce.ts
bd82: yeah but the issue is:
- `path` is a less descriptive name, so I would use `uriPath` which only saves the "dot" (x3)
- Which means I would add 1 line and >20 characters to "save" 20 characters and micro optimize performance.

I think it may be less readable than simply accessing a 1 level deep property thrice.

packages/vscode-dependencies-validation/src/editorChanges.ts
bd82: here I always have the document so I corrected as you suggested.

packages/vscode-dependencies-validation/src/editorChanges.ts
bd82: I tried that earlier, but than the type of `uri` would be `Uri | undefined`

image

So I would need to adjust the code in ways that make less sense, e.g:
- nested `if` inside `if`
- additional condition to check `uri` is undefined

So lacking a "perfect" way I just choose to stick with the simplest one
Can you suggest an alternative?

packages/vscode-dependencies-validation/src/editorChanges.ts
bd82: same as the reply above.
`editor` may be undefined so we may not have a `uri` here...

packages/vscode-dependencies-validation/src/diagnostics/debounce.ts
bd82: I don't think it will be a common edge case, and even if "garbage" gets collected it is unlikely to cause a serious memory leak. But better safe than sorry considering it seems simple to prevent it...

packages/vscode-dependencies-validation/src/editorChanges.ts
slavik-lvovsky: just change the if to ---> if (uri && shouldBeChecked(uri.path))

packages/vscode-dependencies-validation/src/diagnostics/debounce.ts
slavik-lvovsky: Ok. IMHO this one is better: const {path: uriPath} = uri;

PR #168, opened 2022-01-06
feat: add logger to extension

slavik-lvovsky:


packages/vscode-dependencies-validation/src/logger/logger.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Note the use of a getter function so the value would be lazy resolved on each use.
+ * This enables concise and simple consumption of the Logger throughout our extension.
+ */
+export function getLogger(): IVSCodeExtLogger {
+  return logger;
+}
+
+/* istanbul ignore next - ignoring "legacy" missing coverage to enforce all new code to be 100% */
+export function initLogger(context: ExtensionContext): void {
+  const extensionName = "vscode-dependency-validation"; // If the extension name changes, change this too
```
bd82, line 21: i think `context.extension.id` should give you the name

bd82: if not make a **test** to match this const value with what is in the pkg.json

slavik-lvovsky: we already have an output channel with this name.

packages/vscode-dependencies-validation/src/logger/logger.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Note the use of a getter function so the value would be lazy resolved on each use.
+ * This enables concise and simple consumption of the Logger throughout our extension.
+ */
+export function getLogger(): IVSCodeExtLogger {
+  return logger;
+}
+
+/* istanbul ignore next - ignoring "legacy" missing coverage to enforce all new code to be 100% */
```
bd82, line 19: the istanbul ignore comment here should be removed

packages/vscode-dependencies-validation/nyc.config.js
```diff
@@ -3,5 +3,5 @@ module.exports = {
   // No integration or end to end tests currently in this project.
   // `extension.ts` only contains minimal integration code with VSCode.
   // and can be "safely" excluded.
-  exclude: ["src/extension.ts", "src/commands.ts"],
+  exclude: ["src/extension.ts", "src/commands.ts", "src/logger/logger.ts"],
```
bd82, line 6: you may also need to add exclusions for the **root** nyc.config 
so it won't fail when it does the coverage merging

packages/vscode-dependencies-validation/src/logger/logger.ts
```diff
@@ -0,0 +1,40 @@
...
+export function initLogger(
+  context: ExtensionContext,
+  outputChannel: OutputChannel
+): void {
+  const extensionName = context.extension.id; // If the extension name changes, change this too
```
bd82, line 24: the comment `// If the extension name changes, change this too` is not relevant anymore

packages/vscode-dependencies-validation/src/logger/logger.ts
```diff
@@ -0,0 +1,40 @@
...
+  try {
+    logger = configureLogger({
+      extName: extensionName,
+      logPath: context.logPath,
+      logOutputChannel: outputChannel,
+      loggingLevelProp: LOGGING_LEVEL_CONFIG_PROP,
+      sourceLocationProp: SOURCE_TRACKING_CONFIG_PROP,
+      subscriptions: context.subscriptions,
+    });
+  } catch (error) /* istanbul ignore next -- this is complex to test and will give little value */ {
```
bd82, line 34: we are excluding this whole file from coverage so maybe the istanbul comment is redundant

bd82 (approved):
see minor comments

packages/vscode-dependencies-validation/src/logger/logger.ts
slavik-lvovsky: Done.

packages/vscode-dependencies-validation/nyc.config.js
slavik-lvovsky: Done.

packages/vscode-dependencies-validation/src/logger/logger.ts
slavik-lvovsky: Done.

packages/vscode-dependencies-validation/src/logger/logger.ts
slavik-lvovsky: Done.

PR #170, opened 2022-01-10
chore: add trace messages

slavik-lvovsky:


packages/vscode-dependencies-validation/src/diagnostics.ts
```diff
@@ -2,10 +2,13 @@ import type { DiagnosticCollection, Uri } from "vscode";
 import { Range, Diagnostic } from "vscode";
 import { isEmpty } from "lodash";
 import { findDependencyIssues } from "@sap-devx/npm-dependencies-validation";
+import { IChildLogger } from "@vscode-logging/types";
 import { getLogger } from "./logger/logger";
 import { NPM_DEPENDENCY_ISSUES_CODE } from "./constants";
 
-const logger = getLogger().getChildLogger({ label: "diagnostics" });
+export function logger(): IChildLogger {
```
bd82, line 9: why is this exported?

packages/vscode-dependencies-validation/src/logger/logger.ts
```diff
@@ -1,4 +1,4 @@
-import { ExtensionContext, OutputChannel } from "vscode";
+import type { ExtensionContext, OutputChannel } from "vscode";
```
bd82, line 1: 👍

packages/vscode-dependencies-validation/src/diagnostics.ts
bd82: should it be exported via `internal` property if it is for tests purposes?

bd82 (approved):
see one minor in-lined comment

packages/vscode-dependencies-validation/src/diagnostics.ts
slavik-lvovsky: Done

PR #171, opened 2022-01-11
fix: output channel infinite loop

bd82:


bd82:
I think the failing tests should be deleted, I don't see the value in fixing them.
Their TCO is greater than the value provided.

- Given a set of functions was called does not **strongly** prove the unit/feature works.
- If a small refactor happens then the tests need to be adjusted, but these tests have a strong coloration with how a code review would by a human would be done, so we are kinda automating parts of the code review process, but that adds additional overhead.

WDYT?

bd82:
You can fix the tests if you so choose, I'm not sure we will keep them in the long run though...

bd82:
+1

slavik-lvovsky (approved):

PR #175, opened 2022-02-01
refactor: avoid using npm ls

bd82:


packages/npm-dependencies-validation/src/logger.ts
```diff
@@ -1,9 +1,5 @@
 import { OutputChannel } from "./types";
 
-export function print(data: string, outputChannel?: OutputChannel): void {
-  if (outputChannel) {
+export function print(data: string, outputChannel: OutputChannel): void {
     outputChannel.appendLine(data);
-  } else {
```
bd82, line 6: This logger/print is not very generic currently.
And also we don't actually need it right now, so it is simpler to remove it
instead of adding a test case for it...

packages/npm-dependencies-validation/src/types.ts
```diff
@@ -1,19 +1,4 @@
...
-  name: string;
-  version: string;
-} & DependenciesProperties;
-
-export type DependenciesProperties = {
-  [key in DependenciesPropertyName]?: {
-    [key: string]: string;
-  };
-};
+import { PackageJson } from "type-fest";
```
bd82, line 1: "type-fest" is very nice for missing "built-in" types
- https://github.com/sindresorhus/type-fest

packages/vscode-dependencies-validation/src/util.ts
```diff
@@ -22,8 +23,26 @@ export async function findAndFixDepsIssues(
...
+
+/**
+ * Temp "glue" until we display errors on exact position
+ * instead of a single error at the top of the package.json file.
+ */
+export function depIssuesToProblemStrings(issues: DepIssue[]): string[] {
+  const problems = map(issues, (currIssue) => {
+    switch (currIssue.type) {
+      case "mismatch":
+        return "";
```
bd82, line 41: TODO: add the messages here

bd82: done

bd82:
I am merging this as I have another PR which depends on it.
@slavik-lvovsky  Please perform the review async

PR #176, opened 2022-02-02
feat: display diagnostics at correct position

bd82:


bd82:
need to check why it says:
```
editing the appropriate files and running:
│
│     patch-package @types/jest
│
│   If that doesn't work, then it's a bug in patch-package, so please submit a bug
│   report. Thanks!
│     https://github.com/ds300/patch-package/issues
│
│ ---
│ patch-package finished with 1 error(s).
```

packages/vscode-dependencies-validation/src/diagnostics/messages.ts
```diff
@@ -0,0 +1,18 @@
...
+export function depIssueToDiagnosticMsg(issue: DepIssue): string {
+  switch (issue.type) {
+    case "missing":
+      return `The "${issue.name}" package is not installed`;
+    case "mismatch":
+      return (
+        `The "${issue.name}" package installed version ` +
+        `"${issue.actual}", does not match the declared range "${issue.expected}"`
+      );
+    // TODO: do we need a default handler?
```
bd82, line 16: I think we should throw an error in this case

packages/vscode-dependencies-validation/src/util.ts
```diff
@@ -24,28 +24,8 @@ export async function findAndFixDepsIssues(
 ): Promise<void> {
   const { fsPath } = packageJsonUri;
   const depIssues = await findDependencyIssues(fsPath);
-  const problems = depIssuesToProblemStrings(depIssues);
-  if (isEmpty(problems)) return;
-
+  if (isEmpty(depIssues)) {
```
bd82, line 27: jetbrains does not seem to support smart breakpoints within the same line
so i cannot debug it without the block

packages/vscode-dependencies-validation/tsconfig.json
```diff
@@ -5,5 +5,9 @@
     "outDir": "dist",
     "baseUrl": "."
   },
-  "include": ["src/**/*", "test/**/*"]
+  "include": ["src/**/*", "test/**/*"],
+
+  "typeAcquisition": {
```
bd82, line 10: TODO: remove this, its no longer needed

bd82:
TODO: check LGTM alerts
```

    2 for Unused variable, import, function or class
```

packages/vscode-dependencies-validation/tsconfig.json
bd82: done

packages/vscode-dependencies-validation/src/diagnostics/messages.ts
bd82: solved with
- https://github.com/ts-essentials/ts-essentials#exhaustive-switch-cases

packages/vscode-dependencies-validation/patches/@types+jest+27.4.0.patch
```diff
@@ -0,0 +1,35 @@
...
+-declare var afterEach: jest.Lifecycle;
+-declare var describe: jest.Describe;
+-declare var fdescribe: jest.Describe;
+-declare var xdescribe: jest.Describe;
+-declare var it: jest.It;
+-declare var fit: jest.It;
+-declare var xit: jest.It;
+-declare var test: jest.It;
+-declare var xtest: jest.It;
++// declare var beforeAll: jest.Lifecycle;
```
bd82, line 21: conflicts with mocha definitions, but we need this to compile the jest-mock-vscode package...

packages/vscode-dependencies-validation/src/diagnostics/refreshDiagnostics.ts
```diff
@@ -0,0 +1,29 @@
+import type { DiagnosticCollection, Uri } from "vscode";
+import { readFile } from "fs-extra";
+import { findDependencyIssues } from "@sap-devx/npm-dependencies-validation";
+import { convertToDiagnostics } from "./convertToDiagnostics";
+
+// TODO: evaluate if we make a small unit test for this or not? (probably yes... as it does not seem hard)
```
bd82, line 6: TODO: test?

bd82: Wrote a test for this but it requires too much investment compare to the complexity of the code and the chance that this trivial 4 line function would cause a "real" bug...

bd82:
@slavik-lvovsky this should be ready for review

slavik-lvovsky (approved):

PR #180, opened 2022-02-10
Edited.

Paolaufer:
Feel free to contact me to go over my changes.

### DIFF
--- a/packages/vscode-dependencies-validation/README.md
+++ b/packages/vscode-dependencies-validation/README.md
@@ -1,7 +1,7 @@
-# NPM Dependency Validation
+# NPM Dependencies Validation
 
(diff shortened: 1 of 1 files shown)


bd82:
Thanks @Paolaufer 

I applied your 98% of your fixes suggestions here:
- https://github.com/SAP/app-studio-toolkit/pull/181/files

Please note I could not change the title (dependencies --> Dependency) as I would also need
to change the technical name of the component and that is out of scope for this PR.

Also note the markdown gets rendered on github anyhow and not displayed "as-is",
so I'm intentionally adding line breaks so the lines would 
fit in the IDE (normally 80-100 chars per line max), particularly in bullets.

PR #192, opened 2022-05-15
fix: support for identifying LCAP mode also for vscode

EstherPerelman:
### What it does:
Updates the way to identify LCAP mode to be determined by the existence of the LCAP (cockpit) extension,
this should be right both for theia & vscode

- [x]  Tested locally with the `app-studio-toolkit` and `lcap-cockpit` extensions `vsix`

### DIFF
--- a/packages/app-studio-toolkit/src/apis/validateLCAP.ts
+++ b/packages/app-studio-toolkit/src/apis/validateLCAP.ts
@@ -1,19 +1,14 @@
-import { getLogger } from "../logger/logger";
-import { optionalRequire } from "../utils/optional-require";
+import { getLogger } from "@sap/artifact-management";
+import { extensions } from "vscode";
 
-export async function isLCAPEnabled(): Promise<boolean | undefined> {
+export const LACP_EXTENSION_ID = "SAPSE.lcap-cockpit";
+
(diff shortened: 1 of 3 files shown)


NetaMolcho (approved):

packages/app-studio-toolkit/src/apis/validateLCAP.ts
```diff
@@ -1,19 +1,14 @@
...
+  const logger = getLogger().getChildLogger({ label: "isLCAPEnabled" });
 
-  const isLCAPEnabled = await sapPlugin.window.isLCAPEnabled();
-  logger.trace("LCAP enabled successfully received.", { isLCAPEnabled });
+  // LCAP mode is determined by the existence of the LCAP extension
+  const isLCAPEnabled = !!extensions.getExtension(LACP_EXTENSION_ID);
+  logger.trace("LCAP enabled mode", { isLCAPEnabled });
 
-  return isLCAPEnabled as boolean | undefined;
+  return Promise.resolve(isLCAPEnabled);
```
bd82, line 13: this function is **async** so an additional promise (`Promise.resolve`) wrapper is not needed.

A non promise value returned from an async promise will be automatically wrapped
in a resolved promise.

PR #197, opened 2022-09-29
fix: fix fiori header icons

veredcon:
Now it will look like this:

image

Fixing issue:
https://jira.tools.sap/browse/DEVXCORE-2569
https://github.wdf.sap.corp/devx-wing/vscode-ide-extension/issues/111


veredcon:
Now it will look like this:

image

idoprz (approved):

PR #200, opened 2022-12-06
feat: update artiface-management from 1.3.0 to 1.17.2

ThreeWind:
The issue of webpack bundling has been solved in @sap/artifact-management 1.17.2. At the same time, the WorkspaceApi and WorkspaceImpl have been refactored since @sap/artifact-management 1.17.0.

In this pull request, I updated @sap/artifact-management from 1.3.0 to 1.17.2 by removing the unused patch and refactoring serval codes related to workspace to align with the above changes in artifact-management.


packages/app-studio-toolkit/src/project-type/workspace-instance.ts
```diff
@@ -1,14 +1,13 @@
...
-import { workspace } from "vscode";
+import * as vscode from "vscode";
 
 let workspaceAPI: WorkspaceApi;
 
 /**
  * "Factory" for the `WorkspaceApi` "original" instance.
  */
 export function initWorkspaceAPI(): WorkspaceApi {
-  // @ts-expect-error - we "patched" the `WorkspaceImpl` constructor while waiting for official patch.
```
bd82, line 10: 👍

bd82:
@ThreeWind, please sign the CLA: https://github.com/SAP/app-studio-toolkit/pull/200#issuecomment-1338988144

packages/app-studio-toolkit-types/package.json
```diff
@@ -20,7 +20,7 @@
     "compile": "tsc -p ."
   },
   "dependencies": {
-    "@sap/artifact-management-types": "1.3.0",
+    "@sap/artifact-management-types": "1.17.2",
```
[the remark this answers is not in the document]
ThreeWind: Good question. Yes, I agree with you. I will expose the other two methods startWatch() and stopWatch().

bd82:
> @bd82 I have signed the CLA several times, but I do not know why the status is always pending?

I am not sure, but as a SAP employee the CLA is redundant so we can skip it.

bd82:
> Hi @bd82 We still need those the CopyPlugin in webpack.config.js because Project API need those templates to generate the launchpad page and mat.yaml.

Yeah, but we only expose three APIs from `workspaceImpl` and I doubt the project-api is used as a code generator in this flow. So the question is: would the project API run successfully if these resources are "missing"? 
Basically does it lazy load them when needed? or always loads them  on its initialization?

packages/app-studio-toolkit-types/package.json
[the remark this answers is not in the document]
bd82: I believe we intentionally did not expose these start/stop watch methods so one BAS extension would not be able to affect another. (singleton project-api...)

packages/app-studio-toolkit/test/public-api/create-workspace-proxy.spec.ts
```diff
@@ -19,6 +18,8 @@ describe("the `createWorkspaceProxy` utility", () => {
       onWorkspaceChanged(...args: any[]) {
         return args;
       },
+      startWatch() {},
```
bd82, line 21: 1. I don't think we should expose this methods.

packages/app-studio-toolkit/src/public-api/create-workspace-proxy.ts
```diff
@@ -12,6 +12,8 @@ export function createWorkspaceProxy(
     onWorkspaceChanged: (
       ...args: Parameters<WorkspaceApi["onWorkspaceChanged"]>
     ) => workspaceImpl.onWorkspaceChanged(...args),
+    startWatch: () => workspaceImpl.startWatch(),
```
bd82, line 15: i don't think we should expose this methods.
What happens if one BAS extension does "stopWatch"?
I assume it would "stop" watching for **all** BAS extensions
using this singleton project-api object...

ThreeWind:
> > Hi @bd82 We still need those the CopyPlugin in webpack.config.js because Project API need those templates to generate the launchpad page and mat.yaml.
> 
> Yeah, but we only expose three APIs from `workspaceImpl` and I doubt the project-api is used as a code generator in this flow. So the question is: would the project API run successfully if these resources are "missing"? Basically does it lazy load them when needed? or always loads them on its initialization?

Actually, we expose all interfaces of project-api by using the line 'export * as sam from "@sap/artifact-management-types";' in app-studio-toolkit-types, not just the three APIs from WorkspaceImpl. Typically, users use WorkspaceApi to get ProjectApi objects and call their methods to perform all kinds of tasks, such as getProjectInfo(), readItems() or build() etc. If there resources are missing, some of the methods can not run successfully.

packages/app-studio-toolkit/src/public-api/create-workspace-proxy.ts
ThreeWind: I totally agree with you. Thanks for correcting me and I will revert this commit.

packages/app-studio-toolkit/test/public-api/create-workspace-proxy.spec.ts
ThreeWind: Yes, you are right.

ThreeWind:
hi @bd82, would you please review this pull request once more? It cannot be merged without your approval.

bd82:
> Actually, we expose all interfaces of project-api by using the line 'export * as sam from "@sap/artifact-management-types";' in app-studio-toolkit-types, not just the three APIs from WorkspaceImpl. Typically, users use WorkspaceApi to get ProjectApi objects and call their methods to perform all kinds of tasks, such as getProjectInfo(), readItems() or build() etc. If there resources are missing, some of the methods can not run successfully.

Good point which shows where we need to improve the types.
This `sam` type is meant as a design time only object, we do not actually expose it at runtime.
It was added so consumers would be able to access artifact-management **interfaces/types**, e.g: "Item / Module".
I wonder if we can do a "cleaner/smaller" selection of sub-types to expose there.

But that is out of scope of this PR.

PR #201, opened 2022-12-26
chore: update "Last Run" debug icon color on vscode top header  for fiori light quartz theme

EstherPerelman:


alex-gilin (changes requested):
what happens in the patch if the 'lastRunDebugIcon.startForeground' key is lost/missing ? consider to add a test to cover this case

PR #203, opened 2023-01-04
feat: expose project api cli methods from BAS commands

ThreeWind:
In order to call project api commands through the basctl tool in BAS terminal, a VSCode command project-api.command.run is introduced. The VSCode command receives the request from the basctl tool and sends it to project api.

The related backlog item: https://jira.tools.sap/browse/LCAP-2780

### DIFF
--- a/packages/app-studio-toolkit/package.json
+++ b/packages/app-studio-toolkit/package.json
@@ -28,7 +28,7 @@
     "package": "node ./scripts/package-vsix.js"
   },
   "dependencies": {
-    "@sap/artifact-management": "1.17.2",
+    "@sap/artifact-management": "1.18.0",
     "@vscode-logging/wrapper": "1.0.1",
     "lodash": "4.17.21"
   },
@@ -81,6 +81,12 @@
         "label": "SAP Fiori Quartz Light",
         "uiTheme": "vs",
         "path": "./src/themes/light-default-clean.json"
+      }
+    ],
+    "commands": [
+      {
+        "command": "project-api.command.run",
+        "title": "Run a specific command of Project API"
       }
     ]
   },
--- a/packages/app-studio-toolkit/src/project-type/workspace-instance.ts
+++ b/packages/app-studio-toolkit/src/project-type/workspace-instance.ts
@@ -1,7 +1,7 @@
 import { WorkspaceImpl, WorkspaceApi } from "@sap/artifact-management";
 import * as vscode from "vscode";
 import { CommandExecutor } from "@sap/artifact-management";
-import { isPlainObject } from "lodash";
+import { isPlainObject, isString } from "lodash";
 
 let workspaceAPI: WorkspaceApi;
 
@@ -23,7 +23,14 @@ export function initWorkspaceAPI(
           if (isPlainObject(result)) {
             result = JSON.stringify(result, undefined, 2);
           }
-          return !result ? "\n" : (result as string) + "\n";
+
+          if (isString(result)) {
+            return (result ) + "\n";
+          } else if (result === undefined) {
+            return "\n";
+          } else {
+            return "Unsupported type of result: " + typeof result;
+          }
         } catch (error) {
           return (
             `Failed to execute the command ${command} with the error: ${error.message}` +
--- a/packages/app-studio-toolkit/src/project-type/workspace-instance.ts
+++ b/packages/app-studio-toolkit/src/project-type/workspace-instance.ts
@@ -1,6 +1,7 @@
 import { WorkspaceImpl, WorkspaceApi } from "@sap/artifact-management";
 import * as vscode from "vscode";
 import { CommandExecutor } from "@sap/artifact-management";
+import { isPlainObject } from "lodash";
 
 let workspaceAPI: WorkspaceApi;
 
@@ -19,10 +20,10 @@ export function initWorkspaceAPI(
       async (command, ...params) => {
         try {
           let result = await CommandExecutor.execute(command, ...params);
-          if (typeof result !== "string") {
+          if (isPlainObject(result)) {
             result = JSON.stringify(result, undefined, 2);
           }
-          return (result as string) + "\n";
+          return !result ? "\n" : (result as string) + "\n";
         } catch (error) {
           return (
             `Failed to execute the command ${command} with the error: ${error.message}` +
--- a/packages/app-studio-toolkit/src/project-type/workspace-instance.ts
+++ b/packages/app-studio-toolkit/src/project-type/workspace-instance.ts
@@ -1,14 +1,37 @@
 import { WorkspaceImpl, WorkspaceApi } from "@sap/artifact-management";
 import * as vscode from "vscode";
+import { CommandExecutor } from "@sap/artifact-management";
 
 let workspaceAPI: WorkspaceApi;
 
 /**
  * "Factory" for the `WorkspaceApi` "original" instance.
  */
-export function initWorkspaceAPI(): WorkspaceApi {
+export function initWorkspaceAPI(
+  context: vscode.ExtensionContext
+): WorkspaceApi {
   workspaceAPI = new WorkspaceImpl(vscode);
   workspaceAPI.startWatch();
+
+  context.subscriptions.push(
+    vscode.commands.registerCommand(
+      "project-api.command.run",
+      async (command, ...params) => {
+        try {
+          let result = await CommandExecutor.execute(command, ...params);
+          if (typeof result !== "string") {
+            result = JSON.stringify(result, undefined, 2);
+          }
+          return (result as string) + "\n";
+        } catch (error) {
+          return (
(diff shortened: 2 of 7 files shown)


packages/app-studio-toolkit/package.json
```diff
@@ -28,7 +28,7 @@
     "package": "node ./scripts/package-vsix.js"
   },
   "dependencies": {
-    "@sap/artifact-management": "1.17.2",
+    "@sap/artifact-management": "1.18.0",
```
bd82, line 31: should not the `artifact-managment-types` be also updated 
in `app-studio-types` package?

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,122 @@
+import { mockVscode } from "../mockUtil";
+import { ok } from "assert";
```
bd82, line 2: can you please use chai for assertions
to stay consistent with the rest of this project?
`import { expect } from "chai"`

I think it is also a much richer assertion library
- https://www.chaijs.com/api/assert/

bd82: It has a `match` utility on strings
- expect('foobar').to.match(/^foo/);

packages/app-studio-toolkit/package.json
ThreeWind: Thanks for your reminder. I will update artifact-managment-types to 1.18.0.

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
ThreeWind: Yes, it is better choice to use chai here. I will replace assert with chai in this file.

packages/app-studio-toolkit/src/project-type/workspace-instance.ts
```diff
@@ -1,14 +1,37 @@
...
   workspaceAPI = new WorkspaceImpl(vscode);
   workspaceAPI.startWatch();
 
+  context.subscriptions.push(
+    vscode.commands.registerCommand(
+      "project-api.command.run",
+      async (command, ...params) => {
+        try {
+          let result = await CommandExecutor.execute(command, ...params);
+          if (typeof result !== "string") {
```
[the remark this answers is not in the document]
ThreeWind: Actually, the execute result can be just three kinds: string, object and undefined (no return). For the result is undefined, I will change code and add a unit test to support it. 
I agree with you that it helps make logic cleaner by using isPlainObject, so I will make changes based on your suggestion.

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+  it("run BAS command and return STRING result", async () => {
+    const context: any = {
+      subscriptions: {
+        push: () => {},
+      },
+    };
+
+    workspaceInstance.initWorkspaceAPI(context);
+
+    expect(commandParmater).to.be.equal("project-api.command.run");
```
bd82, line 70: ~to.be.equal~ --> to.equal

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    commandExecutorMock
+      .expects("execute")
+      .withExactArgs("testCommand", "testProjectPath")
+      .returns("testResult");
+    const result = await callbackParmater("testCommand", "testProjectPath");
+    expect(result.startsWith("testResult")).to.be.true;
```
bd82, line 77: ~.to.match(/^foo/);~
-->
expect(result).to.match(/^testResult/);

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+    expect(commandParmater).to.be.equal("project-api.command.run");
+
+    commandExecutorMock
+      .expects("execute")
+      .withExactArgs("testCommand", "testProjectPath")
+      .returns({
+        result: "Result is an object",
+      });
+    const result = await callbackParmater("testCommand", "testProjectPath");
+    expect(result.indexOf("Result is an object") >= 0).to.be.true;
```
bd82, line 98: ~expect(result.indexOf("Result is an object") >= 0).to.be.true;~
-->
expect(result).to.be.an('array').that.includes("Result is an object");

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+  it("unsupported command", async () => {
+    const context: any = {
+      subscriptions: {
+        push: () => {},
+      },
+    };
+
+    workspaceInstance.initWorkspaceAPI(context);
+
+    expect(commandParmater).to.be.equal("project-api.command.run");
```
bd82, line 128: ~to.be.equal~ --> to.equal

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    commandExecutorMock
+      .expects("execute")
+      .withExactArgs("unsupportedCommand", "testProjectPath")
+      .throws(new Error("Unsupported command"));
+    const result = await callbackParmater(
+      "unsupportedCommand",
+      "testProjectPath"
+    );
+    expect(result.indexOf("Unsupported command") >= 0).to.be.true;
```
bd82, line 138: expect(result).to.be.an('array').that.includes("Unsupported command");

packages/app-studio-toolkit/src/project-type/workspace-instance.ts
```diff
@@ -1,14 +1,38 @@
...
+  context.subscriptions.push(
+    vscode.commands.registerCommand(
+      "project-api.command.run",
+      async (command, ...params) => {
+        try {
+          let result = await CommandExecutor.execute(command, ...params);
+          if (isPlainObject(result)) {
+            result = JSON.stringify(result, undefined, 2);
+          }
+          return !result ? "\n" : (result as string) + "\n";
```
bd82, line 26: It seems strange that you add a `newline` to the results.
The formatting should be a concern of whoever prints those results out, not your concern...

bd82: e.g maybe they use different newlines terminals?
or print the results in a table without newlines?

bd82: also relying on the `!result` automatic boolean conversion to detect a string is not precise.

e.g: `!"" === true`

you are already using `isPlainObject` from lodash.
So you can also use isString.

If you want to add special handling for undefined
I would make this handling condition explicit:
`if x === undefined`

Otherwise this handler could also be run for other cases (e.g null)

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
```diff
@@ -0,0 +1,140 @@
+import { mockVscode } from "../mockUtil";
+import { expect } from "chai";
```
bd82, line 2: many of your assertions seem to be directly translated from `assert` to `chai`

I left possible  examples how to better use `chai`s rich API
which will result in better error messages in the console when running the tests (for failing tests).

bd82 (changes requested):
see inline comments

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
ThreeWind: The result is string not an array, so I will change the line to expect(result).includes("Result is an object");

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
ThreeWind: The result is string not an array, so I will change the line to expect(result).includes("Unsupported command");

packages/app-studio-toolkit/test/project-type/workspace-instance.spec.ts
ThreeWind: Thanks very much for your advice.
I do not have much experience of using chai. Your corrections about chai make me learn a lot. I will make changes based on your advice.

packages/app-studio-toolkit/src/project-type/workspace-instance.ts
bd82: The formatting with newline still seems a little strange to me, but this is your flow and if you are sure
this is the optimal approach we can keep it.

PR #313, opened 2024-05-20
feat: create a unique `basctl` server socket for each extension host

alex-gilin:


packages/app-studio-toolkit/src/basctlServer/basctlServer.ts
```diff
@@ -64,7 +66,9 @@ function createBasctlServer() {
   }
 }
 
-export function startBasctlServer() {
+export function startBasctlServer(context: vscode.ExtensionContext) {
+  // add the socket file name to the environment variables so that the basctl client can use it
```
veredcon, line 70: Maybe add a link to the relevant place where it is used? e.g:
https://github.wdf.sap.corp/devx-wing/bas-cli/blob/master/src/socketClient.js#L4

packages/app-studio-toolkit/src/basctlServer/basctlServer.ts
```diff
@@ -5,8 +5,10 @@ import * as _ from "lodash";
 import { _performAction } from "../actions/performer";
 import { ActionsFactory } from "../actions/actionsFactory";
 import { getLogger } from "../logger/logger";
+import { v4 as uuidv4 } from "uuid";
 
-const SOCKETFILE = "/extbin/basctlSocket";
+// generate a unique socket file name on each run
+const SOCKETFILE = `/extbin/basctl-${uuidv4()}.sock`;
```
veredcon, line 11: So there is one unique socketfile per each app toolkit activation, right? So for multiple windows in each one there will be a different socketfile

veredcon (approved):

packages/app-studio-toolkit/src/basctlServer/basctlServer.ts
alex-gilin: it is an OSS and the link you mentioned will be unreachable externally

packages/app-studio-toolkit/src/basctlServer/basctlServer.ts
alex-gilin: exactly. unique socket connection for each extension host

PR #395, opened 2025-03-03
feat: add telemetry reporter api

veredcon:


packages/app-studio-toolkit/src/telemetry/telemetryInit.ts
```diff
@@ -0,0 +1,21 @@
+import { setup, defaultClient } from "applicationinsights";
+import { APPINSIGHTS_CONNECTION_STRING } from "./constants";
+
+export function initializeTelemetry() {
+  setup(APPINSIGHTS_CONNECTION_STRING);
+
+  // Configure sample rate. 100 is the default and means all collected data will be sent to the Application Insights service
+  // If you want to enable sampling to reduce the amount of data, set the samplingPercentage. 0 means nothing will be sent.
+  defaultClient.config.samplingPercentage = 100;
```
manwithafox, line 9: What is this samplingPercentage setting? Who can change it? As this seems to be a global setting for all telemetry players it basically means if one player reports to much and then reduces the percentage, messages of all other players are also reduced. 
In our vscode-cds extension we have already reduced the data by reporting collected statistics (with only average numbers) for most items. We send those every 5mins. Actually I would not be happy if one of those requests would get filtered out due to a lowered sampling rate.

manwithafox (commented):
Looks promising, but pls see my comments.

packages/app-studio-toolkit/src/telemetry/telemetryInit.ts
```diff
@@ -0,0 +1,21 @@
+import { setup, defaultClient } from "applicationinsights";
+import { APPINSIGHTS_CONNECTION_STRING } from "./constants";
+
+export function initializeTelemetry() {
```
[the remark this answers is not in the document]
veredcon: Good point Christian! :) I fixed it and now it is singleton-like initialization, as currently we need one instance for all reporters.

packages/app-studio-toolkit/src/telemetry/telemetryInit.ts
veredcon: `samplingPercentage` cannot be changed by the extension's reporter and effect other players. It is initialized globally/centrally in one app insights client for all reporters. In case we, as BAS telemetry FW, decides to reduce the sampling percentage, then yes it will effect all players. 

If we would store different app insights client per extension (as it was before by fix now) we could in theory offer `samplingPercentage` optional parameter to the `getTelemetryReporter` API so each reporter can define its own percentage, but we are also not thrilled with that since we would lose the control to reduce it as a platform.

For example, recently we found that the analytics might affect cost traffic and increase overall costs, so we need to maintain the ability to control it as a platform when and if it will be required.

packages/app-studio-toolkit/src/telemetry/basTelemetryClient.ts
```diff
@@ -0,0 +1,110 @@
...
+  getDataCenterParam,
+  getBASMode,
+  getHashedUser,
+} from "./utils";
+import { getExtensionRunPlatform } from "../utils/bas-utils";
+interface TelemetryProperties {
+  [key: string]: string | boolean;
+}
+
+interface TelemetryMeasurements {
```
joergmann, line 17: I'm wondering why **TelemetryMeasurements** and **TelemetryProperties** are separated? In our API we have one bag and split it apart based on type before calling your API

joergmann (commented):
Maybe add a simple extension example using the telemetry API? 
Other than that: LGTM

packages/app-studio-toolkit/src/telemetry/telemetryInit.ts
manwithafox: I fully understand your reasoning. Wouldn't it be better to dynamically adjust the percentage PER player?
In the end it is the amount of telemetry data/events that should be controllable.
Why should a player that sends less events be reduced only because some other player floods the telemetry backend?

packages/app-studio-toolkit/src/telemetry/basTelemetryClient.ts
veredcon: Hi Joerg, we wanted our API to be aligned with the underlying `trackEvent` API of app insights which gets event with this structure: https://github.com/microsoft/ApplicationInsights-node.js/blob/main/src/declarations/contracts/telemetryTypes/eventTelemetry.ts#L7

joergmann: I guessed something like that. Anyway, we can keep our structure and map

NetaMolcho (approved):

PR #396, opened 2025-03-04
fix: validation issue in monorepo

vbangelov:


nyc.config.js
```diff
@@ -10,10 +10,10 @@ module.exports = {
     "packages/vscode-deps-upgrade-tool/src/logger.ts",
   ],
   //   - https://reflectoring.io/100-percent-test-coverage/
-  branches: 100,
-  lines: 100,
-  functions: 100,
-  statements: 100,
+  branches: 90,
```
vbangelov, line 13: Reduced it temporary because I want to see if the approach to resolving the issue is correct. Will increase it back to 100 once sure that the solution will work.

packages/npm-dependencies-validation/src/depIssuesFinder.ts
```diff
@@ -161,3 +185,95 @@ async function validateMismatchDep(opts: {
...
+    console.error(`Error checking if ${depPkgJsonPath} exists: ${error}`);
+    return false;
+  }
+}
+
+async function getDepInRootNodeModules(
+  depPkgJsonPath: string,
+  depName: string
+): Promise<string> {
+  const projectRoot = await getProjectRoot(depPkgJsonPath);
```
bd82, line 239: It seems you call `getProjectRoot()` for individual each dep being checked for mismatch.
And `getProjectRoot()` is recursive and scans all the way to root of file system.

This seems highly inefficient.
A more efficient logical approach would be to:
1. first understand the project structure (once).
2. Only secondly scan for mismatched deps.

This would require a larger refactoring of the code.
But it cannot be avoided, because there are many types of mono-repo structures / package managers
- yarn pnp / zero installs (this has custom loader without `node_modules` folder)
  - pnpm also supports Plug And Play structure.
- https://bun.sh/package-manager (Not sure what is its structure).

packages/npm-dependencies-validation/src/depIssuesFinder.ts
```diff
@@ -161,3 +185,95 @@ async function validateMismatchDep(opts: {
     return NO_ISSUES;
   }
 }
+
+async function isMonorepo(absPath: string): Promise<boolean> {
+  const packageJsonPath = resolve(absPath, "package.json");
+  const lernaJsonPath = resolve(absPath, "lerna.json");
+  const pnpmWorkspaceYamlPath = resolve(absPath, "pnpm-workspace.yaml");
+
+  // Check for lerna.json or pnpm-workspace.yaml
```
bd82, line 194: its not enough to check for `lerna.json` or `pnpm-workspace.yaml` in some parent.
The sub-package from which we have arrived must also be included inside the workspace definition...

This is another indication the logical flow to validate this should first "understand" the whole project structure
and only then perform the validations for all the package.json files.

bd82 (changes requested):
It seems as though you are adding a flag to modify the behavior and check a **single** additional `node_modules` folder for the missing dependency.

I don't think this approach would scale.
1. There are different types of mono-repos structures.
2. The mono-repo is defined by the root metadata, so the implementation logic should start from there, not from a sub-package.
3. The logic to detect if we are inside a mono-repo is incomplete (no bi-directional references).

Basically I suspect adding mono-repo support would require a larger refactor for the code. Otherwise future changes will introduce more and more technical debt.

Also note the implementation is very inefficient in terms of file system access

bd82:
It may also be best to defer the lookup of the packages to the detect package manager (yarn/pnpm/npm/other).
Otherwise we will  end up (incorrectly) **re-implementing** logic of package managers, but never be fully alligned.

PR #489, opened 2026-05-12
chore: migrating `webide-client-tools` from a separate repository int…

alex-gilin:
…o this one as a package


package.json
```diff
@@ -31,14 +31,6 @@
       "@commitlint/config-conventional"
     ]
   },
-  "lint-staged": {
```
bd82, line 34: why would you disable the entire lint-staged configuration for this whole mono repo?

tsconfig.json
```diff
@@ -21,6 +21,9 @@
     {
       "path": "./packages/vscode-deps-upgrade-tool"
     },
+    {
```
bd82, line 24: this ts-config is a helper for same topic mono repo. (tsc --watch)
I don't think its relevant for web-ide-client-tools

bd82: but if it works you can keep it...

packages/webide-client-tools/package.json
```diff
@@ -0,0 +1,95 @@
...
+  "scripts": {
+    "ci": "npm-run-all format:validate type_check lint test",
+    "clean": "shx rm -rf dist .nyc_output docs/web/html_docs",
+    "lint": "eslint src/**/*.js test/**/*.js test_integration/**/*.js",
+    "test": "jest",
+    "test_integration": "npm-run-all init_example execute_integration_tests unlink_example",
+    "init_example": "node scripts/init_example.js",
+    "execute_integration_tests": "jest --maxWorkers 1 --config test_integration/jest.config.json",
+    "unlink_example": "cd ./example && npm unlink @sap-webide/webide-client-tools",
+    "format:fix": "prettier --write \"**/*.@(js|ts|json|md)\" --ignore-path=.gitignore",
```
bd82: formatting is handled by the root package.json of this monorepo

packages/webide-client-tools/package.json
```diff
@@ -0,0 +1,95 @@
...
+    "eslint-config-airbnb-base": "15.0.0",
+    "eslint-plugin-eslint-comments": "3.2.0",
+    "eslint-plugin-import": "2.25.3",
+    "jest": "27.4.2",
+    "prettier": "2.5.0"
+  },
+  "scripts": {
+    "ci": "npm-run-all format:validate type_check lint test",
+    "clean": "shx rm -rf dist .nyc_output docs/web/html_docs",
+    "lint": "eslint src/**/*.js test/**/*.js test_integration/**/*.js",
```
bd82: linting is handled by the root of this repo.
I don't think you need to merge the package specific eslint plugins to the root config
just make sure the package specific eslint flow runs (as is) on this sub-package.

packages/webide-client-tools/README.md
```diff
@@ -0,0 +1,71 @@
+[![Continuous Integration](https://github.com/SAP/app-studio-toolkit/actions/workflows/ci.yml/badge.svg)](https://github.com/SAP/app-studio-toolkit/actions/workflows/ci.yml)
```
bd82: I don't think you need all these sub-badges in a sub-package
check what other sub-packages are using in this monorepo

.changeset/gentle-chicken-slide.md
```diff
@@ -0,0 +1,38 @@
+---
+"@sap-webide/webide-client-tools": major
```
bd82, line 2: do you really need to release a new version because you moved this project to a new mono-repo?

bd82: you might just need to do the minimum changes so the projects runs in a modern build
and also resolve security scanning issues (renovate)

So modernize only the **design** time of the project, not modernize the **runtime** and release a new major version.

If can release a new version if and when it is required (hopefully never...)

bd82:
I cannot fully review this, when you both add +138 files in a **single commit**
which **contains** both the:
1. migration of code
2. modernization of code.

I will need to use AI to extracts and visualize the diffs of the **modernization** part...

package.json
bd82: ah i see it was extracted to lint-staged.config.js

alex-gilin: Not at all. On the contrary: the `lint-staged.config.js` file I added *is* that very configuration. It is picked up by the `lint-staged` utility (which is triggered by the `pre-commit` hook from Husky with every Git commit).

packages/webide-client-tools/package.json
alex-gilin: Processing is still performed as root. This script was part of the package even before it was included in the mono repo. But could be removed - agree.

packages/webide-client-tools/package.json
alex-gilin: Similar to the case with the `format` script.

.changeset/gentle-chicken-slide.md
alex-gilin: The minimum required version of `Node.js` is now 20 (`engines: ">=20"`), `webpack` has been upgraded from version 4 to 5, and bugs have been fixed and dependencies updated—all of which constitute sufficient grounds for a major release.

packages/webide-client-tools/package.json
alex-gilin: removed

packages/webide-client-tools/README.md
alex-gilin: removed

bd82 (approved):
I cannot fully review the modernizing parts as its mixed it.
But I can approve the migration and what I noticed from the modernizing parts.

PR #496, opened 2026-05-13
ci: use app for short lived token

bd82:


dellagustin-sap:
Hello @bd82 , I'm not an expert in GitHub Actions workflows, but I did a reviewed this PR and got some questions.
If I understood correctly, the token produced by your GitHub App is only needed for the step `ChangeSets Release or PR creation`, yet it is being used on `actions/checkout`, and thus available to all the subsequent steps.

I wonder if you could achieve the same results by restricting the usage of `steps.app-token.outputs.token` only to when it is needed.

I have proposed a refactored version with #514 , on top of this PR.
Note that I'm not 100% sure that would work, you would need to test it.

PR #503, opened 2026-05-28
ci: integrate yeoman-ui into app-studio-toolkit monorepo 

alex-gilin:


package.json
```diff
@@ -18,8 +18,8 @@
     "format:fix": "prettier --write \"**/*.@(js|ts|json|md)\" --ignore-path=.gitignore",
     "format:validate": "prettier --check \"**/*.@(js|ts|json|md)\" --ignore-path=.gitignore",
     "preinstall": "npx only-allow pnpm",
-    "legal:copy-license": "pnpm -r exec -- shx cp -r ../../LICENSES LICENSES",
-    "legal:copy-reuse": "pnpm -r exec -- shx cp -r ../../.reuse .reuse",
+    "legal:copy-license": "shx cp -r LICENSES projects/yeoman-ui/ && pnpm -r exec -- shx cp -r ../../LICENSES LICENSES",
```
bd82: modifying this root script for every single new "project" we integrate into this monorepo will not scale.

**Looks like we need a custom nodejs script for this...**

You can try using something like:
`pnpm ls -r --depth -1` to list the packages and paths.

AI suggested this manner for calling in a script
```javascript
const { execSync } = require('child_process');

function getMonorepoPackages() {
  try {
    // Run the pnpm command with the --json flag
    const stdout = execSync('pnpm ls -r --depth -1 --json', { encoding: 'utf8' });
    
    // Parse the JSON string into an array of package objects
    return JSON.parse(stdout);
  } catch (error) {
    console.error('Failed to list pnpm workspaces:', error.message);
    return [];
  }
}

// Execute and inspect the results
const packages = getMonorepoPackages();
console.log(packages);

```

pnpm-workspace.yaml
```diff
@@ -1,10 +1,22 @@
 packages:
   - 'packages/*'
+  - 'projects/yeoman-ui/packages/*'
```
bd82: 1. would it accept a more general pattern? e.g.: 'projects/*/packages/*'
2. Why do we need the `packages/` in the path?
    - There is no longer a root package.json under `projects/yeoman-ui/package.json`
    - So is the `packages` sub-folder still needed?
    - This is subjective, I don't have a clear answer, maybe its good to keep this format to allow easy extraction from the mono-repo in the future....

pnpm-workspace.yaml
```diff
@@ -1,10 +1,22 @@
...
 allowBuilds:
   '@vscode/vsce-sign': true
   bufferutil: true
+  core-js: true
+  ejs: true
   es5-ext: true
   esbuild: true
   keytar: true
   utf-8-validate: true
+overrides:
```
bd82: We **must not** use global overrides.
as this monorepo will contain a great many packages.

Use specific per sub-package overides instead
- https://pnpm.io/settings#overrides

Image

tsconfig.json
```diff
@@ -32,6 +32,12 @@
     },
     {
       "path": "./examples/vscode-using-upgrade-tool"
+    },
```
bd82: I don't think we need to update this tsconfig.json
Its for local dev flows, but that flow is no longer relevant when this repo
becomes a multi projects monorepo

projects/yeoman-ui/tsconfig.base.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "strict": false
+  },
+  "exclude": ["node_modules", "dist"]
```
bd82: this exclude is a little odd.

It might help you if you define this yeoman-ui **project** tsconfig **separately**
from the **monorepo**.

We can defer the question of setting up a singular central tsconfig for the future.

projects/yeoman-ui/.eslintrc.js
```diff
@@ -0,0 +1,74 @@
+module.exports = {
```
bd82: There is already an eslint config at the root of the project
and unlike the tsconfig.json I think there is great value in having a central consistent shared config among all projects.

In addition the newer versions of ESLint (v9+) do not seem to support per folder config.


I would recommend trying to merge the configs.
The result does not have to be perfect, we can improve it later.
But having a shared skeleton eslint config is the way to go for this multi-project monorepo.

Also note this is a very simple config with only a few popular plugins
and some disabling of rules, so it should not be too hard (hopefully)

projects/yeoman-ui/CHANGELOG.md
```diff
@@ -0,0 +1,552 @@
+# Change Log
```
bd82, line 1: this changelog will no longer be updated because the changesets does not 
recognize this project folder as a package (no package.json).

So add a line at the top mentioning this is the historic changelog
and any newer changes will appear in the packages themselves.

projects/yeoman-ui/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,86 @@
+# SAP Open Source Code of Conduct
```
bd82: we do not need a code of conduct at the project level.
This already appears at the top of the mono-repo

projects/yeoman-ui/CONTRIBUTING.md
```diff
@@ -0,0 +1,165 @@
+# Contributing to an SAP Open Source Project
```
bd82: This file should only include things specific to `yeoman-ui` project.

Things like:
- General Remarks
- Developer Certificate of Origin (DCO)
- Legal
- others...

Are already in the root mono-repo CONTRIBUTING.md

projects/yeoman-ui/CONTRIBUTING_USING_GENAI.md
```diff
@@ -0,0 +1,12 @@
+# Guideline for AI-generated code contributions to SAP Open Source Software Projects
```
bd82, line 1: this file is at the root of the mono-repo alraedy

projects/yeoman-ui/README.md
```diff
@@ -0,0 +1,89 @@
...
+- [Application Wizard UI](./packages/frontend) The Application Wizard as a standalone vue.js application.
+- [Yeoman generator sample](./packages/generator-foodq) Sample yeoman generator to show usages and test the platform.
+- [![npm-yeoman-ui-types][npm-yeoman-ui-types-image]][npm-yeoman-ui-types-url] [@sap-devx/yeoman-ui-types](./packages/types) Type signatures supposed to be used in the yeoman generators.
+
+[npm-yeoman-ui-types-image]: https://img.shields.io/npm/v/@sap-devx/yeoman-ui-types.svg
+[npm-yeoman-ui-types-url]: https://www.npmjs.com/package/@sap-devx/yeoman-ui-types
+
+## Support
+
+To get more help, support, and information please open a github [issue](https://github.com/SAP/app-studio-toolkit/issues).
```
bd82: lets create issue labels per project

projects/yeoman-ui/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Report an Issue
```
bd82, line 27: These sections:

- `Report an Issue`
- `Issue handling process`

Would be duplicated between all projects in this mono-repo.
Should be extracted to main README.md and linked.

Assuming  you think it belongs in main README.md.
Maybe it belongs in main CONTRIBUTING.md?

The important thing is to avoid the duplication

app-studio-toolkit.code-workspace
```diff
@@ -0,0 +1,13 @@
+{
```
bd82, line 1: did you test this `.code-workspace` works even when no at the root of the repo?

bd82 (changes requested):
See inline comments

I started with reviewing the top level structure and integration into the monorepo

projects/yeoman-ui/.eslintrc.js
alex-gilin: done

projects/yeoman-ui/CODE_OF_CONDUCT.md
alex-gilin: removed

projects/yeoman-ui/CONTRIBUTING_USING_GENAI.md
alex-gilin: removed

projects/yeoman-ui/CONTRIBUTING.md
alex-gilin: done

projects/yeoman-ui/README.md
alex-gilin: done

projects/yeoman-ui/README.md
alex-gilin: done

projects/yeoman-ui/tsconfig.base.json
alex-gilin: removed

app-studio-toolkit.code-workspace
alex-gilin: Why is this needed? The purpose of this is to ensure that `.vscode` subconfigurations are available when loading a repository into `.code-workspace`, rather than simply opening the folder (where `.vscode` subconfigurations are skipped).

package.json
alex-gilin: done

pnpm-workspace.yaml
alex-gilin: adjusted

pnpm-workspace.yaml
alex-gilin: refactored

tsconfig.json
alex-gilin: removed

projects/yeoman-ui/CHANGELOG.md
alex-gilin: done

scripts/legal-copy.js
```diff
@@ -0,0 +1,65 @@
+/**
```
bd82, line 1: This script is perfectly fine for its use case.

Just be aware that if you need top level **async**
you can use *.**m**js file which supports it.
For future scripts...

app-studio-toolkit.code-workspace
bd82: my bad i thought it was related to github codespaces

projects/yeoman-ui/README.md
```diff
@@ -22,23 +22,11 @@ This npm [mono-repo][mono-repo] currently contains:
 
 ## Support
 
-To get more help, support, and information please open a github [issue](https://github.com/SAP/app-studio-toolkit/issues).
+To get more help, support, and information please open a GitHub [issue](https://github.com/SAP/app-studio-toolkit/issues/new?labels=project%3Ayeoman-ui).
```
bd82, line 25: 👍

projects/yeoman-ui/CONTRIBUTING_USING_GENAI.md
bd82: you seem to have moved the file into the sub-package.
So deleting from root and adding in the new yeoman-ui project root

image

bd82 (changes requested):
only one tiny thing
it seems the "contribution with genai" markdown file was moved from the root
into the new project imported into the monorepo.

bd82 (approved):

PR #513, opened 2026-05-31
fix(webide-client-tools): increase bundleFeature test timeout to 20s

alex-gilin:
10s was too tight for webpack bundling on CI runners, causing a timeout that left stale dist folders and cascaded into a second failure (expected 1 timestamp folder, found 11).

### DIFF
--- a/packages/webide-client-tools/test/bundling_spec.js
+++ b/packages/webide-client-tools/test/bundling_spec.js
@@ -378,7 +378,7 @@ describe("The, Exported bundling APIs", () => {
 
   function defineBundleFeatureSpecs(bundler) {
     describe(`bundleFeature API - ${bundler}`, () => {
-      const bundleTestTimeout = 10000
+      const bundleTestTimeout = 20000
       const distFolder = path.resolve(__dirname, "dist")
       const expectedJSOutputFile = `${distFolder}/config-preload.js`
       const expectedMetadataOutputFile = `${distFolder}/config-preload.json`


packages/webide-client-tools/test/bundling_spec.js
```diff
@@ -378,7 +378,7 @@ describe("The, Exported bundling APIs", () => {
 
   function defineBundleFeatureSpecs(bundler) {
     describe(`bundleFeature API - ${bundler}`, () => {
-      const bundleTestTimeout = 10000
+      const bundleTestTimeout = 60000
```
bd82: this is a little risky as we will combine many many packages into this monorepo.
If you set 5-6 tests a 60 seconds timeout, at some point in the future some regression
could cause them to really use most of that time for valid run.

But we have to consider also the whole mono-repo total tests runtime.
I don't think its a problem now because all the tests (everything) together pass in 100 seconds

Image

So, maybe use a smaller buffer? e.g.: 20 seconds.

Do you have a link to build log where the failure occurred ?
You can also reduce possible conflict / race conditions with the tests
by making each test case use a **different** folder
It seems they are all using the same folder right now.

bd82 (changes requested):
see review comment

packages/webide-client-tools/test/bundling_spec.js
alex-gilin: used smaller buffer - 20 sec

bd82 (approved):

PR #514, opened 2026-06-02
ci: Review PR #496 - Reduce usage of token with elevated permissions

dellagustin-sap:
This commit refactors the `release.yml` workflow to to reduce the usage of the token generated with the GitHub app.

actions/checkout is now using the default token and not persisting credentials, as they are not needed (e.g., _Upload VSIX to releases_ receives the token through the environment.

The token is now generated just before it is needed.


bd82:
Thank you @dellagustin-sap

bd82 (approved):

PR #526, opened 2026-06-17
ci: skip pr comment workflow for forked pulled requests

bd82:
As forked prs lack the required permissions.
And the repo's security may be (slightly) compromised even if implement mitigations
as recommended here: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

For example, one must remember that the published artifacts by the forked PR (*.vsix) can themselves
contain malicious code and should not be freely installed without oversight.

Additionally keeping this PR Comments workflow for forked PRs is not a high priority
as its an optional helper flow, and forked PRs are not common on this repo.


alex-gilin (approved):
LGTM

PR #540, opened 2026-06-22
chore: bump typescript to 5.7.3 and eslint tooling

ilya-taupeka:


bd82 (changes requested):
We need to avoid disabling `skipLibCheck` 
this is too brute force and disables for all libs.

Can you upgrade the failing libs so they don't break?
Failing that can you exclude specific libs that break during compilaton?

I got this answer from GPT 5.5

```
skipLibCheck was added because the last commit upgrades TypeScript from 4.5.2 to 5.7.3, and TS 5.7 starts failing on dependency declaration files when lib checking is enabled.
I reproduced one of the failures by forcing lib checks back on:
node_modules/.pnpm/type-fest@2.11.1/node_modules/type-fest/source/get.d.ts(127,37): error TS2344
node_modules/.pnpm/type-fest@2.11.1/node_modules/type-fest/source/get.d.ts(128,36): error TS2344
So the change is effectively a build unblocker: skip checking .d.ts files from dependencies such as type-fest@2.11.1, while still type-checking this repo’s source.
Why many files: this repo has many independently compiled package/example tsconfig.json files. Since tsconfig.base.json does not set skipLibCheck, each leaf project that can hit the dependency declaration issue needed it locally. Some packages already had it; the commit adds it to the ones that were missing it.
The cleaner alternatives would be either upgrading/replacing the incompatible dependency declarations, especially type-fest@2.11.1, or setting skipLibCheck once in tsconfig.base.json. The commit chose the narrower local suppression.

```

bd82 (approved):

PR #541, opened 2026-06-22
feat: support multiple yeoman-env versions

ilya-taupeka:
- update yeoman-environment to 6.1.0 version
- update yeoman-generator to 8.2.2 version
- add patch loader to fix CJS bundling
- add backward compatibility with old generators


projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
+    path.join(meta.packagePath, "node_modules"),
+    path.join(path.dirname(meta.resolved), "node_modules"),
+  ];
+
+  for (const root of nodeModulesRoots) {
+    const pkgJson = path.join(root, "yeoman-generator", "package.json");
+
+    try {
+      const version: string =
+        JSON.parse(fs.readFileSync(pkgJson, "utf8")).version ?? "";
```
bd82, line 36: don't use `fs.readFileSync` use the **async** versions instead, in this case `isLegacyGenerator()` is being called from an async method, so doing this is safe and easier on nodejs single threaded event loop.

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
 import * as _ from "lodash";
-import { homedir } from "os";
 import * as path from "path";
-import { existsSync } from "fs";
-import { isWin32, NpmCommand } from "./npm";
+import * as fs from "fs";
```
bd82, line 3: use named imports, don't import entire packages.
this makes bundling easier and more minimalist

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
 const NAMESPACE = "namespace";
 
+// Detects whether a generator requires yeoman-environment v3 (legacy) or v6 (modern)
+//
+// Detection strategy:
+// 1. Private node_modules copy: read yeoman-generator/package.json from the generator's
+//    own node_modules subtree — version major ≤ 5 → legacy
+// 2. Declared dep range: read the generator's own package.json dependencies field —
+//    covers generators that webpack-bundled their deps (no node_modules subfolder)
+// 3. Bundle scan: read the generator's resolved entry file and look for `env.runLoop` —
```
bd82, line 22: # 3 logic is not robust

1. Bundling with certain configurations can produce multiple entry points.
2. bundling can produce different levels of minimificaiton of name mangling, so text search on the bundled.js is not safe.

If `3.` algo is necessary for the solution it may not be solvable in a dynamic manner and we might need to. go back to include/exclude list of legacy generators.

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
+//    the exact property name checked by yeoman-generator ≤ v5
+function isLegacyGenerator(meta: LookupGeneratorMeta): boolean {
+  // --- Check 1: private node_modules copy ---
+  const nodeModulesRoots = [
+    path.join(meta.packagePath, "node_modules"),
+    path.join(path.dirname(meta.resolved), "node_modules"),
```
bd82, line 28: This second path with `meta.resolved` does not seem to be a possible (valid) yeoman generator structure.

Did you encounter some legacy generator that required this lookup?

## See AI answer:

**Short Answer**
Probably not in normal Yeoman usage.

**Normal Layout**
For a typical generator package, dependencies live under the generator package root:

```txt
node_modules/generator-foo/
  package.json
  node_modules/yeoman-generator/
  generators/app/index.js
```

So this path is the meaningful one:

```ts
path.join(meta.packagePath, "node_modules")
```

**Second Layout**
The second path:

```ts
path.join(path.dirname(meta.resolved), "node_modules")
```

would imply something like:

```txt
node_modules/generator-foo/
  generators/app/
    index.js
    node_modules/yeoman-generator/
```

That is not a known Yeoman legacy convention. Older Yeoman generators commonly used `app/index.js`, `generators/app/index.js`, or `lib/generators/app/index.js`, but dependencies were still package-level dependencies.

**Conclusion**
I’d read the second path as defensive or over-broad handling, not a real legacy Yeoman pattern.

A more accurate approach would be Node resolution from the generator file:

```ts
require.resolve("yeoman-generator/package.json", {
  paths: [path.dirname(meta.resolved)],
});
```

That would naturally check nearby `node_modules` folders and walk upward according to Node’s normal resolution rules.

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
 import { IChildLogger } from "@vscode-logging/logger";
 import { getClassLogger } from "../logger/logger-wrapper";
 
 const GENERATOR = "generator-";
 const NAMESPACE = "namespace";
 
+// Detects whether a generator requires yeoman-environment v3 (legacy) or v6 (modern)
+//
+// Detection strategy:
+// 1. Private node_modules copy: read yeoman-generator/package.json from the generator's
```
bd82, line 18: This makes an assumption on only supporting limited sub-set of valid node_modules structures.

See the full algorithim npm uses for lookup here:
- https://docs.npmjs.com/cli/v6/commands/npm-install#algorithm

It is likely better to use `import.meta.resolve` or `require.resolve` for full compatability with any current / future `node_modules` structure, e.g.: dev-spaces recently switched from `pnpm` to `bun` for global installs

- depending on your runtime env.
- Note that these APIs do not look in global paths by default
(but you can probably pass the global paths to search in via their APIs)

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
+    path.join(path.dirname(meta.resolved), "node_modules"),
+  ];
+
+  for (const root of nodeModulesRoots) {
+    const pkgJson = path.join(root, "yeoman-generator", "package.json");
+
+    try {
+      const version: string =
+        JSON.parse(fs.readFileSync(pkgJson, "utf8")).version ?? "";
+      const major = parseInt(version.split(".")[0], 10);
```
bd82, line 37: This is trivial, and yet you should still use semVer for parsing
semantic versions

https://www.npmjs.com/package/semver

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
+import Environment, { createEnv } from "yeoman-environment";
+import type { EnvironmentOptions } from "yeoman-environment";
+import type { LookupGeneratorMeta } from "@yeoman/types";
 import { IChildLogger } from "@vscode-logging/logger";
 import { getClassLogger } from "../logger/logger-wrapper";
 
 const GENERATOR = "generator-";
 const NAMESPACE = "namespace";
 
+// Detects whether a generator requires yeoman-environment v3 (legacy) or v6 (modern)
```
bd82, line 15: this comment does not match the actual implementation.
it seems legacy === <= 5.0.0

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
+    const pkgJson = path.join(root, "yeoman-generator", "package.json");
+
+    try {
+      const version: string =
+        JSON.parse(fs.readFileSync(pkgJson, "utf8")).version ?? "";
+      const major = parseInt(version.split(".")[0], 10);
+
+      return !isNaN(major) && major <= 5;
+    } catch {
+      // not present at this root, try next check
```
bd82, line 41: some kind of logging would help debug issues here

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
+  // --- Check 2: declared dependency range in generator's package.json ---
+  // Handles generators that bundled deps and may not have a node_modules subfolder
+  try {
+    const genPkg = JSON.parse(
+      fs.readFileSync(path.join(meta.packagePath, "package.json"), "utf8")
+    );
+    const range: string =
+      (genPkg.dependencies ?? {})["yeoman-generator"] ??
+      (genPkg.devDependencies ?? {})["yeoman-generator"] ??
+      (genPkg.peerDependencies ?? {})["yeoman-generator"] ??
```
bd82, line 54: This resolution logic can gets quite complicated


1. Different npm clients verisons can behave differently in regards to peerDeps automatically being installed
2. devDeps do not automatically get installed, but they might list a version which is later bundled.

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
+  } catch {
+    // generator has no readable package.json, try next check
+  }
+
+  // --- Check 3: scan resolved entry file for env.runLoop usage ---
+  // yeoman-generator ≤ v5 checks `env.runLoop` in its constructor — this property name
+  // is structural and survives minification. Generators that fully
+  // bundle their deps (no node_modules, no dep in package.json) still carry this pattern
+  try {
+    const bundle = fs.readFileSync(meta.resolved, "utf8");
```
bd82, line 74: async fs operations

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -1,24 +1,100 @@
...
+
+    return bundle.includes("env.runLoop");
```
bd82, line 76: as mentioned, its hard to prove it survives minification for all possible current and future minifiers / bundlers, even if the name is global it could be written as an equivalent 

```javascript
env["runLoop"] = function()...
``

From experience parsing JS source code for `reflection` logic  is very fragile.

File: projects/yeoman-ui/packages/backend/src/utils/yeoman-env-all-patch-loader.js [hunk 12]
```diff
@@ -0,0 +1,97 @@
+"use strict";
```
bd82, line 1: this *.js should not be in `src/`
`src/` is for runtime *.ts

this script is part of the design time bundle script, it should be in `/scripts`
or `/scripts/webpaack/`

projects/yeoman-ui/packages/backend/src/utils/yeoman-env-all-patch-loader.js
```diff
@@ -0,0 +1,97 @@
...
+
+// Webpack loader that patches all yeoman-environment dist files for CJS bundling.
+//
+// yeoman-environment v6 is a pure ESM package. When webpack bundles it into a CJS
```
bd82, line 5: newest vscode should support esm inside exteension host.
perhaps it should be tested to avoid the bundling complexity

projects/yeoman-ui/packages/backend/src/utils/yeoman-env-all-patch-loader.js
```diff
@@ -0,0 +1,97 @@
...
+// output (libraryTarget: "commonjs2"), several patterns break:
```
bd82, line 6: this **bundle time patching** seems extremely fragile consider future verison of yeoman-env

projects/yeoman-ui/packages/backend/test/yeomanui.spec.ts
```diff
@@ -1635,10 +1640,11 @@ describe("yeomanui unit test", () => {
       expect(
         doGeneratorDoneSpy.calledWith(
           true,
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
           _.get(
             yeomanUi,
             "uiOptions.messages.artifact_with_name_generated",
-            (_: string) => ""
+            (_a: string) => ""
```
bd82, line 1647: many changes here which seem unrelated to the yeoman-env legacy versions topic

projects/yeoman-ui/packages/backend/package.json
```diff
@@ -29,7 +29,7 @@
   "publisher": "SAPOS",
   "main": "./dist/extension",
   "scripts": {
-    "bundle": "webpack --mode production",
+    "bundle": "pnpm --filter yeoman-env-compat bundle && shx cp ../yeoman-env-compat/dist/index.js ./dist/yeoman-env-compat.js && webpack --mode production",
```
bd82, line 32: don't access sibling parent via `../`
declare a devDependency to it via `workspace:*` version in package.json.

and it will appear in this package's local `node_modules` dir.

the declaration of dependency is also important for upcoming tools like turbo repo which run the build in parallel
- https://turborepo.dev/
- which i intend to introduce to this repo to mitigate increasing build times as more packages are added to this mono-reo

projects/yeoman-ui/packages/backend/webpack.config.js
```diff
@@ -27,7 +27,17 @@ const config = {
   },
   externals: {
     vscode: "commonjs vscode",
+    // yeoman-environment and yeoman-generator are ESM-only packages (type: "module").
+    // They cannot be loaded via CommonJS require() on Node < 22.12 (ERR_REQUIRE_ESM).
```
bd82, line 31: do we care about version < 22.12?
The whole 22 major line reaches EOL in ~10 months
and if needed we can control nodejs version on BAS

bd82: There seems to be a-lot of complexity here which would likely be reduced significantly if we can run ESM vscode extension more natively

projects/yeoman-ui/packages/yeoman-env-compat/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "yeoman-env-compat",
+  "displayName": "yeoman-env-compat",
+  "version": "1.25.0",
+  "private": true,
+  "description": "Thin wrapper that re-exports yeoman-environment v3 and yeoman-generator v4 for backward-compatible legacy generators",
```
bd82, line 6: I thought only the old env is needed for legacy generators?

projects/yeoman-ui/packages/yeoman-env-compat/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "yeoman-env-compat",
```
bd82, line 2: should you encode the versions in the package name?
e.g.:

- `yeoman-env-v3` -> `yeoman-env-compat`

e.g.: what happens when you try to create another similar package with yeoman-env v4 in future?

- yeoman-env-compat
- yeoman-env-compat2
- yeoman-env-compat3

projects/yeoman-ui/packages/yeoman-env-compat/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
bd82, line 1: why is a tsconfig needed if there is no `*.ts?`

projects/yeoman-ui/packages/yeoman-env-compat/webpack.config.js
```diff
@@ -0,0 +1,143 @@
+"use strict";
+
+const path = require("path");
```
bd82, line 3: We might be able to trade off `*.vsix` size to reduce this bundling complexity, basically the *.vsix can include
`node_modules` but I don't know if its possible for "partial" node_modules (subset).

bd82 (commented):
We might need to go back to the drawing board.

1. Needed env resolution mode seem fragile and I'm not sure this is resolvable.
2. Bundling logic of newer `yo-env` includes actual **patches** which is also fragile particularly because newer `yo-env` is a "moving target" (patches will break...).
3. Old `yo-env` is also none trivial, but this might be okay as it can be considered a more static version.

## To discuss

### Dynamic Env Choice logic
1. is this even feasible in a maintainable / robust manner.
2. Should we implement an `includeList` approach to support specific legacy generators instead?

### Dynamic Env (legacy / new) Running in VSCode.

One is ESM and one is CJS. This automatically creates none trivial bundling issues.
Should we instead have two Yeoman-UI extensions existing at the same time?
- Yeoman-ui-legacy (CJS VSCode Extension)
- Yeoman-ui-Modern (ESM VSCode Extension)

That way each *.vsix can handle the bundling in a minimal way most compatiable with its runtime env...

Also the legacy code won't prevent the `modern` variant from moving forward in terms of tooling / libraries due to CJS backward compatibility.

Please setup a followup meeting to discuss these

projects/yeoman-ui/packages/yeoman-env-compat/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "directory": "projects/yeoman-ui/packages/yeoman-env-compat"
+  },
+  "scripts": {
+    "bundle": "webpack --config webpack.config.js",
+    "compile": "tsc",
+    "clean": "shx rm -rf ./dist"
+  },
+  "dependencies": {
+    "yeoman-environment": "3.19.3",
+    "yeoman-generator": "4.13.0"
```
bd82, line 23: does this env bundling require generator inside as well?

ilya-taupeka:
irrelevant

PR #546, opened 2026-06-29
feat: migrate yeoman-ui project to esm

ilya-taupeka:


projects/yeoman-ui/packages/backend/package.json
```diff
@@ -27,21 +27,22 @@
     "name": "SAP SE"
   },
   "publisher": "SAPOS",
-  "main": "./dist/extension",
+  "type": "module",
```
bd82, line 30: Can we verify that the local dev websocket flows still work after this change?

With `"type": "module"`, the existing `ws:run` / `ws:egRun` scripts execute `dist/src/webSocketServer/*.js` as native ESM:

```json
"ws:egRun": "node ./dist/src/webSocketServer/exploregens.js",
"ws:run": "node ./dist/src/webSocketServer/youi.js"
```

If these scripts are still supported, they may need updates for native ESM compatibility. For example, depending on how the output is intended to run, the scripts may need to point at `.cjs` files or another bundled/CommonJS entrypoint instead of the emitted `.js` files.

projects/yeoman-ui/packages/backend/tsconfig.json
```diff
@@ -13,10 +13,21 @@
     "baseUrl": ".",
     "sourceMap": true,
     "target": "es6",
+    "module": "esnext",
+    "moduleResolution": "bundler",
```
bd82: This `bundler` resolution seems risky.
It assumes everything is bundled before execution
but that is not true for tests which use `ts-node`

So we could be creating a wider gap between production code and test code behavior

projects/yeoman-ui/packages/backend/test/utils/workspaceFile.spec.ts
```diff
@@ -29,98 +27,77 @@ describe("extension unit test", () => {
...
-      // Expect writeFileSync to be called with the new workspace file path and content
-      fsMock
-        .expects("writeFileSync")
-        .withArgs(expectedWsFilePath, JSON.stringify(fileContent));
-      uriMock.expects("file").withArgs(expectedWsFilePath);
-
-      WorkspaceFile.createWsWithUri(folderConfig); // Pass the FolderUriConfig here
+      // Let the real filesystem handle existsSync - cleanup written files after
+      expect(() => WorkspaceFile.createWsWithUri(folderConfig)).to.not.throw();
+
```
bd82: These tests now write to the real `~/projects` directory via `Constants.HOMEDIR_PROJECTS` and later delete `workspace*.code-workspace` files there. That can remove a developer's real workspace file and makes CI depend on host state. Please run this against a sandbox/temp directory or stub the filesystem so cleanup only removes files created by the test.

projects/yeoman-ui/packages/backend/src/logger/logger-wrapper.ts
```diff
@@ -123,3 +123,15 @@ export function createExtensionLoggerAndSubscribeToLogSettingsChanges(
   // Subscribe to Logger settings changes.
   listenToLogSettingsChanges(context);
 }
+
+export function _setTestLogger(mockLogger: any): void {
```
bd82: Do we need these test-only exports? The tests could initialize the logger through `createExtensionLoggerAndSubscribeToLogSettingsChanges` using a mocked VS Code context/API. That keeps the production module API smaller and exercises the real initialization path instead of mutating the module-scoped singleton directly.

If we do keep test-only helpers, please group them under an explicit internal namespace/const, for example `internal` or `internalTests`, rather than exporting them as top-level production APIs.

projects/yeoman-ui/packages/backend/src/extension.ts
```diff
@@ -16,8 +16,8 @@ export function activate(context: ExtensionContext) {
   void import("./utils/env");
 
   try {
-    createExtensionLoggerAndSubscribeToLogSettingsChanges(context);
-    AnalyticsWrapper.createTracker(context);
+    loggerApi.createExtensionLoggerAndSubscribeToLogSettingsChanges(context);
+    void AnalyticsWrapper.createTracker(context);
```
bd82: I saw you changed this to be async, you can change the activate method to be async too if you want to "await" for `createTracker`

projects/yeoman-ui/packages/backend/src/usage-report/usage-analytics-wrapper.ts
```diff
@@ -31,12 +32,11 @@ export class AnalyticsWrapper {
     return BASClientFactory.getBASTelemetryClient();
   }
 
-  public static createTracker(context: ExtensionContext): void {
+  public static async createTracker(context: ExtensionContext): Promise<void> {
```
bd82, line 35: the intent here was to avoid sync FS access?

This is important but not really in the scope of the desired change.
No need to remove this... but in the future when the PR is 40+ files its best to limit the scope...

projects/yeoman-ui/packages/backend/src/extension.ts
```diff
@@ -1,5 +1,5 @@
 import { ExtensionContext, window, WebviewPanel } from "vscode";
-import { createExtensionLoggerAndSubscribeToLogSettingsChanges } from "./logger/logger-wrapper";
+import { loggerApi } from "./logger/logger-wrapper";
```
bd82: instead of marking "loggerApi" const/namespace as the `productive` api part
you can mark/wrap the test/internal only apis with a namespace.

Productive first makes more sense to me...

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -107,10 +110,10 @@ class EnvUtil {
       generatorName = `${GENERATOR}${genShortName}`;
     }
 
-    const keys = Object.keys(require.cache);
+    const keys = Object.keys(_require.cache);
     for (const key of keys) {
       if (key.includes(generatorName)) {
```
bd82, line 115: This cache invalidation only affects generators loaded through CommonJS `require`. If ESM generators are now supported, this will not unload ESM generator modules between runs because native `import()` uses a separate module cache. Is that acceptable, or do we need a documented limitation/test for ESM generators being cached across runs?

- Or do we need to implement cache unloading for ESM generators as well to be consistent?
- Why do the cache unloading exist in the first place?

projects/yeoman-ui/packages/backend/src/utils/vscodeProxy.ts
```diff
@@ -1,18 +1,23 @@
...
+
+const { set } = lodash;
+
+const _require = createRequire(import.meta.url);
 
 export const getVscode = () => {
   try {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    return require("vscode");
+    return _require("vscode");
```
bd82: if we are moving to an ESM native vscode extension
why are we still mocking VSCode via CJS `require` ?

projects/yeoman-ui/packages/backend/test/panels/YeomanUIPanel.spec.ts
```diff
@@ -19,37 +19,43 @@ describe("YeomanUIPanel unit test", () => {
   let npmUtilsMock: SinonMock;
   let windowMock: SinonMock;
   let commandsMock: SinonMock;
-  let loggerWrapperMock: SinonMock;
   let panel: YeomanUIPanel.YeomanUIPanel;
   let setWebviewPanelStub: SinonStub;
   let createWebviewPanelStub: SinonStub;
   let trackerWrapperMock: SinonMock;
 
+  const mockLogger: any = {
```
bd82: looks like this mock logger can be created by a shared test utils function

projects/yeoman-ui/packages/backend/test/extCommands.spec.ts
```diff
@@ -8,29 +8,38 @@ import { ExtCommands } from "../src/extCommands";
...
   let sandbox: SinonSandbox;
   let windowMock: SinonMock;
-  let loggerWrapperMock: SinonMock;
 
   const testContext: any = {
     subscriptions: [],
     extensionPath: "testExtensionpath",
   };
 
+  const mockLogger: any = {
```
bd82: shared utils to create mock logger (as mentioned before)

projects/yeoman-ui/packages/backend/test/utils/workspaceFile.spec.ts
bd82: Options:

1. Use a nested repo folder that is also in `.gitignore`
2. use temp folder

but don't mutate HOME dir.
very risky

projects/yeoman-ui/packages/backend/webpack.config.cjs
```diff
@@ -21,7 +21,7 @@ const config = {
   output: {
     // the bundle is stored in the 'dist' folder (check package.json), 📖 -> https://webpack.js.org/configuration/output/
     path: path.resolve(__dirname, "dist"),
-    filename: "extension.js",
+    filename: "extension.cjs",
```
bd82: we need to figure out if we can run in ESM natively or not

projects/yeoman-ui/packages/backend/test/vscode-youi-events.spec.ts
```diff
@@ -512,9 +515,6 @@ describe("vscode-youi-events unit test", () => {
         .resolves();
       workspaceMock.expects("updateWorkspaceFolders").withArgs(0, null);
 
-      fsMock.expects("existsSync").returns(false);
```
bd82, line 515: i don't think these mocks and assertions have a-lot of value
but why were they removed without an equivalent replacement?

projects/yeoman-ui/packages/backend/test/mockUtil.ts
```diff
@@ -1,9 +1,10 @@
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import * as mocha from "mocha";
 import { getVscodeMock } from "../src/utils/vscodeProxy";
+import { createRequire } from "module";
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const Module = require("module");
+const _require = createRequire(import.meta.url);
```
bd82: if we are supposed to be running in ESM env, why do we need to mock CJS require("vscode")?

projects/yeoman-ui/packages/backend/test/loaderHooks.mjs
```diff
@@ -0,0 +1,79 @@
+import { createRequire } from "module";
```
bd82: This logic looks quite complex:
And it might **hide** real runtime compatibility problems because of special loading during tests.

We can also consider if we need ts-node to run the tests, seems closer to production to run the *.js or *.mjs compiled files.

AI summarized this as:
```
1. Forces @sap-devx/yeoman-ui-types to resolve to its ESM build.
2. Provides an ESM mock module for vscode during tests.
3. Shims named ESM imports from CJS packages like lodash and @vscode-logging/logger.
```

1. Types should not be imported at runtime, is this a ts-node problem?
2. `vscode` import mock might be needed due to our mock heavy legacy testing approach
3. For lodash we don't have to use named imports, also we can use `lodash-es` if needed.
4. For @vscode-logging/logger, we can use default import instead

```typescript
import logger from "@vscode-logging/logger";
import type {
  getExtensionLoggerOpts,
  IChildLogger,
  IVSCodeExtLogger,
  LogLevel,
} from "@vscode-logging/logger";

const { getExtensionLogger } = logger;
```

bd82 (changes requested):
Main thing is that its still not native ESM running VSCode extension.
Lets pair program this tomororw

projects/yeoman-ui/packages/backend/test/vscode-youi-events.spec.ts
ilya-taupeka: Not removed without replacement — they were superseded by:
```
sandbox.stub(WorkspaceFile, "createWsWithPath").returns(wsFileMockUri);
sandbox.stub(WorkspaceFile, "createWsWithUri").returns(wsFileMockUri);
```
These stubs intercept at a higher level (the WorkspaceFile utility boundary) instead of mocking raw fs calls, which stopped working in ESM (was an error like "Cannot read some property of undefined")

projects/yeoman-ui/packages/backend/src/utils/env.ts
bd82: this logic upgrade to support esm cache invalidation might only be relevant in the next topic/pr of introducing yo-env 6+

projects/yeoman-ui/packages/backend/test/mockUtil.ts
bd82: The confusion / complexity

## Productive Code
*.ts -> *.js (ESM) -> bundle (CJS)

## Test Code
- *.ts directly (via ts-node)
- But has alot special handling for CJS (require), so is it running in ESM or CJS during tests?

projects/yeoman-ui/packages/backend/package.json
bd82: at a minimum you probably need to rename these to *.cjs files
if they don't run in ESM env

projects/yeoman-ui/packages/backend/test/loaderHooks.mjs
ilya-taupeka: loaderHooks.mjs is simplified

projects/yeoman-ui/packages/backend/src/utils/vscodeProxy.ts
```diff
@@ -1,18 +1,12 @@
...
-    return require("vscode");
-  } catch (error) {
-    return undefined;
-  }
-};
+const { set } = lodash;
 
-const filename: string = require?.main?.filename;
-const _isInTest = filename?.includes(join("node_modules", "mocha"));
+const _isInTest = process.argv.some((arg) =>
```
bd82, line 17: I think this file is only used in tests.
If so you may not need to replace the logic to detect `isInTest` but can instead assume its always in test context.

This might be legacy leftover

projects/yeoman-ui/packages/backend/src/exploregens.ts
```diff
@@ -1,10 +1,10 @@
...
-import { IChildLogger } from "@vscode-logging/logger";
-import { IRpc } from "@sap-devx/webview-rpc/out.ext/rpc-common";
+import _ from "lodash";
+import type { IChildLogger } from "@vscode-logging/logger";
+import type { IRpc } from "@sap-devx/webview-rpc/out.ext/rpc-common";
 import { NpmCommand, PackagesData } from "./utils/npm";
 import messages from "./exploreGensMessages";
 import { Env, GeneratorData } from "./utils/env";
-import { vscode } from "./utils/vscodeProxy";
+import * as vscode from "vscode";
```
bd82: his PR broke that dev flow, assuming ws:egRun is still expected to work.
Before the PR:
- src/exploregens.ts imported vscode from ./utils/vscodeProxy.
- src/webSocketServer/exploregens.ts also imported vscode from ../utils/vscodeProxy.
- vscodeProxy.ts did:
return require("vscode");
inside a try/catch, and fell back to getVscodeMock() when plain Node could not resolve vscode.
So in direct Node mode, no real vscode module was required. It used the mock.
After the PR:
- src/exploregens.ts imports directly:
import * as vscode from "vscode";
- ws:egRun still runs plain Node:
"ws:egRun": "node ./dist/src/webSocketServer/exploregens.js"
- There is no test loader/shim in that flow.
- There is no runtime vscode package dependency, only @types/vscode.
So Node should fail resolving "vscode".

projects/yeoman-ui/packages/backend/src/yeomanui.ts
```diff
@@ -231,7 +231,7 @@ export class YeomanUI {
 
       const options = {
         logger: this.logger.getChildLogger({ label: generatorNamespace }),
-        vscode: getVscode(), // TODO: remove this temporary workaround once a better solution is found,
```
bd82, line 234: this might be related to dev-flows (mocked backend for ui)

projects/yeoman-ui/packages/backend/package.json
bd82: its looks like these local dev flows are tightly bound to the productive code, and also
very important for mocking backend vs UI on local browser.

So *.cjs is not enough, this flow would likely need to work in ESM env

projects/yeoman-ui/packages/backend/tsconfig.json
bd82: I think this comment is more true now that we realize the developer script flow is very important and running in pure nodejs env.

I think we should use something similar to:

Image

and solve the compile issues, they are likely most/all related to missing file suffixes in ESM import statements

projects/yeoman-ui/packages/backend/src/utils/vscodeProxy.ts
bd82: why is this comment resolved?

bd82: I am also guessing this proxy is not used everywhere instead of `vscode`
so, some sources will import CJS vscode and some ESM vscode?
Could this cause strange edge cases?

bd82: maybe it should be implemented with some kind of **dynamic import** instead?

projects/yeoman-ui/packages/backend/src/webSocketServer/youi.ts
```diff
@@ -34,7 +34,11 @@ class YeomanUIWebSocketServer {
     wss.on("connection", (ws) => {
       console.log("new ws connection");
       const childLogger: IChildLogger = getConsoleWarnLogger();
-      this.rpc = new RpcExtensionWebSockets(ws, childLogger);
+      // Cast: rpc-extension-ws.d.ts types its ctor param via
+      // `import * as WebSocket from "ws"`, which under node16 resolution
+      // is not structurally interchangeable with the WebSocket class
+      // instance emitted by `on("connection")`.
+      this.rpc = new RpcExtensionWebSockets(ws as any, childLogger);
```
bd82, line 41: And this is a pure types problem? not a runtime problem?

projects/yeoman-ui/packages/backend/.mocharc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "require": ["ts-node/register/transpile-only", "source-map-support/register"],
+  "loader": ["ts-node/esm"],
```
bd82, line 2: 👍

projects/yeoman-ui/packages/types/package.json
```diff
@@ -24,8 +24,10 @@
   "scripts": {
     "ci": "npm-run-all clean compile",
     "clean": "shx rm -rf ./dist",
-    "compile": "tsc -b ./tsconfig.json ./tsconfig.cjs.json ./tsconfig.esm.json ./tsconfig.types.json",
-    "compile:watch": "tsc -watch"
+    "compile": "npm-run-all compile:tsc esm:mark",
+    "compile:tsc": "tsc -b ./tsconfig.json ./tsconfig.cjs.json ./tsconfig.esm.json ./tsconfig.types.json",
+    "compile:watch": "tsc -watch",
+    "esm:mark": "node -e \"require('fs').writeFileSync('./dist/esm/package.json','{\\\"type\\\":\\\"module\\\"}')\""
```
bd82: this seems a bit hacky,
if you need these *.d.ts files under `/esm/` folder to be handled as modules.
you can try modifying the `tsconfig.esm.json`  to produce `*.d.mts` (**m**ts) files.
instead of create a new `package.json` with the type module marker.

projects/yeoman-ui/packages/backend/package.json
```diff
@@ -207,10 +208,11 @@
     "@types/sinon": "^17.0.3",
     "@types/vscode": "^1.50.0",
     "@types/webpack-env": "^1.16.2",
-    "@types/ws": "^7.4.5",
+    "@types/ws": "^8.5.14",
```
bd82, line 211: 👍

bd82 (changes requested):
the only major thing is the dynamic CJS import of vscode in the vscode proxy.
could it be replaced with dynamic ESM import?

projects/yeoman-ui/packages/backend/src/webSocketServer/youi.ts
ilya-taupeka: Yes, pure types, not a runtime problem

projects/yeoman-ui/packages/backend/src/utils/vscodeProxy.ts
bd82: not relevant because its not only in tests

projects/yeoman-ui/packages/types/src/index.mts
```diff
@@ -0,0 +1,224 @@
+export abstract class AppWizard {
```
bd82: this looks like code duplication.

is there a textual difference between the *.cts and *.mts type definitions?
There must be patterns in JavaScript to avoid this.

If the same APIs are exposed do we really need two separate type defs?

bd82 (changes requested):
looks good, we just need to avoid copy/pasting the type definitions

bd82 (approved):

PR #547, opened 2026-06-30
chore: migrate vscode-mta-tools into this mono-repo

TirottaSAP:


projects/vscode-mta-tools/package.json
```diff
@@ -0,0 +1,239 @@
...
+    "@types/tmp": "^0.2.1",
+    "copy-webpack-plugin": "9.0.1",
+    "mock-spawn": "0.2.6",
+    "source-map-support": "0.5.19",
+    "ts-loader": "8.0.16",
+    "tmp": "^0.2.1",
+    "vscode-test": "1.5.2"
+  },
+  "prettier": {
+    "endOfLine": "lf"
```
bd82: prettier settings are in the root of this monorepo.
can probably remove this

pnpm-workspace.yaml
```diff
@@ -9,5 +10,6 @@ allowBuilds:
   ejs: true
   es5-ext: true
   esbuild: true
+  husky: false
```
bd82: how is husky related to this change?

projects/vscode-mta-tools/scripts/package-vsix.js
```diff
@@ -0,0 +1,15 @@
+const { packageCommand } = require("vsce/out/package");
+const { resolve } = require("path");
+
+packageCommand({
+  cwd: resolve(__dirname, ".."),
+  packagePath: undefined,
+  baseContentUrl: undefined,
+  baseImagesUrl: undefined,
+  useYarn: true,
```
bd82, line 12: `useYarn` is suspicious because this repo  uses `PNPM`
I think you should change the *.vsix packaging to work more similiarly to other *.vsix in this repo
or better yet, align with the latest version of the *.vsix packaging tool

- https://www.npmjs.com/package/@vscode/vsce

### These are the details I got from GPT 5.5

useYarn: true is an option passed to vsce’s internal packageCommand.
It means: force vsce to use its Yarn dependency-discovery path instead of npm when collecting files for the .vsix.
In vsce@1.100.1, that specifically affects:
- vscode:prepublish: would run with yarn run vscode:prepublish if that script exists.
- Dependency collection: uses yarn list --prod --json if a yarn.lock exists.
For this PR, the important review point is: this repo uses pnpm, not Yarn. With useYarn: true and no yarn.lock, vsce effectively only starts from the extension folder itself and does not discover production dependencies via npm/pnpm. That is probably okay only if vscode-mta-tools is fully bundled into dist/extension.js, which the .vscodeignore suggests.
So I’d read it as a copied legacy packaging setting from the existing package scripts, not a pnpm-aware setting. If the extension needs runtime files from node_modules, this is risky. If everything is webpack-bundled, it is mostly harmless but misleading.

projects/vscode-mta-tools/.eslintrc.js
```diff
@@ -0,0 +1,67 @@
+module.exports = {
```
bd82: is this .eslintrc.js respected by eslint running from the root?
I think the newest eslint does not support recursive nested configs.
With Yeoman-UI the eslint settings were merged into the root eslint config

projects/vscode-mta-tools/.gitignore
```diff
@@ -0,0 +1,84 @@
+# root files for https://reuse.software
```
bd82: much if not all of this .gitignored is covered by the root `.gitignore` file.
its easy to check locally by:
1. running the build to produce all the artifacts
2. deleting this `.gitignore`
3. `git status` --> are there now unexpected files in git working tree?

projects/vscode-mta-tools/.vscodeignore
```diff
@@ -0,0 +1,17 @@
+# Exclude everything by default (handles pnpm symlinks and hoisted node_modules)
```
bd82: while its probably a good practice they way you changed it to exclude everything first.
you are unnecessarily increasing the scope of this PR by modifying the `.vscodeignore`

## Originally
Image

projects/vscode-mta-tools/CHANGELOG.md
```diff
@@ -0,0 +1,130 @@
+## [1.4.10](https://github.com/SAP/vscode-mta-tools/compare/v1.4.9...v1.4.10) (2023-5-30)
```
bd82: I believe you will need to rename `CHANGELOG.md` --> `CHANGELOG.old.md`
I think not doing this caused problems with release a new version for yeoman-ui (previously migrated)
Please inspect what was done for yeoman-ui

projects/vscode-mta-tools/LICENSE
```diff
@@ -0,0 +1,202 @@
+                                 Apache License
```
bd82: I think the license gets copied from the root, don't need to duplicate it

projects/vscode-mta-tools/README.md
```diff
@@ -0,0 +1,44 @@
+[![CircleCI](https://circleci.com/gh/SAP/vscode-mta-tools.svg?style=svg)](https://circleci.com/gh/SAP/vscode-mta-tools)
```
bd82: please review the README.md

1. some of the badges are no longer relevant for example the `circleci` build status
2. some of the instructions are duplicated in the root (support / license)

Also check what was done with yeoman-ui, I believe a new tag was created for marking issues as related to the relevant project...

bd82 (changes requested):

bd82 (dismissed):

bd82 (approved):

PR #577, opened 2026-07-31
fix(yeoman-ui): sub-generator on wrong yeoman-environment error

ilya-taupeka:


projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -57,12 +62,20 @@ class EnvUtil {
...
         Constants.ENV_INCOMPATIBILITY_MESSAGE_PREFIX
       ) ?? false
     );
   }
 
+  private isV3RuntimeIncompatibilityError(error: unknown): boolean {
+    const message = (error as Error)?.message ?? "";
+    return (
+      message.includes("requires yeoman-environment") ||
+      message.includes("object is not extensible")
```
bd82, line 96: "object is not extensible" is not a yeoman specific error.
Is this due to ESM/CJS compatibility? or?

In what scenario do we get this error?

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -57,12 +62,20 @@ class EnvUtil {
...
+      (error as Error)?.message?.includes(
         Constants.ENV_INCOMPATIBILITY_MESSAGE_PREFIX
       ) ?? false
     );
   }
 
+  private isV3RuntimeIncompatibilityError(error: unknown): boolean {
+    const message = (error as Error)?.message ?? "";
+    return (
+      message.includes("requires yeoman-environment") ||
```
bd82, line 95: does the error `requires yeoman-environment` include more details?
as in the version required?

projects/yeoman-ui/packages/backend/src/utils/env.ts
bd82: Only tests use createEnvAndGen() now

Sounds like it should be removed

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -229,6 +230,91 @@ class EnvUtil {
...
+
+  private async prepareAndRun(
+    env: Environment,
+    gen: any,
+    adapter: any,
+    prepare: PrepareEnvGen
+  ): Promise<void> {
+    adapter?.resetSignal?.();
+    await prepare(env, gen);
+    await Promise.resolve(env.runGenerator(gen));
```
bd82: I think This:

`await Promise.resolve(env.runGenerator(gen));`
is effectively the same as:
`await env.runGenerator(gen);`

projects/yeoman-ui/packages/backend/test/youi-adapter.spec.ts
```diff
@@ -128,4 +128,41 @@ describe("YouiAdapter", () => {
...
+
+    it("progress() runs the provided step function", async () => {
+      const adapter: any = new YouiAdapter(youiEvents, outputChannel);
+      const result = await adapter.progress(() => "done");
+      expect(result).to.equal("done");
+    });
+
+    it("onIdle() resolves", async () => {
+      const adapter: any = new YouiAdapter(youiEvents, outputChannel);
+      await adapter.onIdle();
```
bd82: very strange without assertions, maybe add an assertio so this test is less odd?

` expect(await adapter.onIdle()).to.equal(undefined)`

projects/yeoman-ui/packages/backend/test/yeomanui.spec.ts
```diff
@@ -240,18 +240,16 @@ describe("yeomanui unit test", () => {
...
-          env: {
-            on: () => "",
-            runGenerator: () => "",
-          },
-          gen: {
-            destinationRoot: () => "",
-            on: () => "",
-          },
-        });
+        .callsFake(
```
bd82, line 245: you are now calling `createRunGen()`
but not verifying (asserting) in any way the generator actually ran.
at most its verifying no errors where thrown.

If there are other tests that verify end to end flow with running the generator.
Perhaps these two tests can be removed instead of refactored.

projects/yeoman-ui/packages/backend/src/youi-adapter.ts
bd82, line 72: not really in the scope of this PR, but still interesting: 
Does the prompt adapter work for both v3 and v6?
I understand the signature changed between these versions

projects/yeoman-ui/packages/backend/test/utils/env.spec.ts
```diff
@@ -197,4 +204,283 @@ describe("Env.createEnvAndGen()", () => {
...
+      .stub(Env as any, "loadLegacyV3Compat")
+      .returns(require("yeoman-env-v3"));
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it("routes a legacy top generator to v3 and runs its composed sub-generator to completion", async function () {
+    this.timeout(15000);
```
bd82: are 15 seconds really needed?

bd82 (changes requested):
Missing scenarios in fixture tests
 
- a v6 top generator composing another v6 generator
- a v3 generator composing a v6-only generator --> expected error?
- a v6 generator composing a v3-only generator --> expected error?
- a composed sub-generator failure path

projects/yeoman-ui/packages/backend/src/utils/env.ts
ilya-taupeka: not a yeoman-specific string, it's an ESM/CJS incompatibility: when an ESM generator is instantiated on the v3 environment, v3 tries to write a resolved property onto the module's frozen namespace object and throws

projects/yeoman-ui/packages/backend/src/utils/env.ts
ilya-taupeka: the full message includes the required and current versions, e.g. "This generator requires yeoman-environment at least 4.0.0-rc.0, current version is 3.19.3, try reinstalling latest version of 'yo' or use '--ignore-version-check' option". It's yeoman-generator v7+'s explicit version guard. Template is: `This generator (${this._namespace}) requires ${packageDependency} at least ${versionToCheck}, current version is ${currentVersion}, try reinstalling latest version of 'yo' or use '--ignore-version-check' option`, where, packageDependency is a hardcoded "yeoman-environment", and so "requires yeoman-environment" never changes

projects/yeoman-ui/packages/backend/src/youi-adapter.ts
ilya-taupeka: it works for both. Both runtimes call adapter.prompt(questions, secondArg) — legacy (yeoman-generator v3/v5) passes a callback function as the 2nd arg, modern (v6/v8) passes an initialAnswers object. YouiAdapter.prompt branches on isFunction(cb): it only invokes the 2nd arg when it's actually a function, otherwise it just returns the prompt result

projects/yeoman-ui/packages/backend/src/utils/env.ts
bd82: can we use the details in this error message for env selection?
is the message format stable between different yeoman-generator versions?

WDYT?

bd82: maybe write a short comment that for future multi version support (with more versions)
we may be able to use this error message to select the right version faster

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -166,67 +179,89 @@ class EnvUtil {
...
-    this.logger?.debug(
-      `routing generator ${genNamespace} to default yeoman-environment v6`
-    );
 
+    this.unloadGeneratorModules(genNamespace);
+    let v3EnvGen: EnvGen | undefined;
     try {
-      const v6Env: Environment = this.createEnvInstance(
-        { sharedOptions: { forwardErrorToEnvironment: true } as any },
+      v3EnvGen = this.createLegacyV3EnvAndGen(
```
bd82, line 230: please comment that logic that we try lower version
first the support generators composition scenario
by trying lower version first we will find the lowest overlapping version that both support.

projects/yeoman-ui/packages/backend/test/fixtures/generator-compose-v6-sub/generators/app/index.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
+
+// A modern (yeoman-generator v6) sub-generator. Requires the v6 runtime
+const Generator = require("yeoman-generator");
```
bd82: do explicit import of the `yeoman-generator` version
by using an alias, just like the other tests.

Otherwise future updates to the project may break the test or worse break the test intent

projects/yeoman-ui/packages/backend/test/fixtures/generator-compose-v6-only-sub/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "generator-compose-v6-only-sub",
```
bd82: generally please review the fixture names.
Try to go for explicit / clear names.

For example:
- gen-v8-env6
- gen-v6-env6
- compose-parent-gen-v6-env-v6-sub-... 

1. maybe this one is best to ask the AI to refactor?
2. maybe use `positive/negative` sub-folders to split up the fixtures.

projects/yeoman-ui/packages/backend/test/fixtures/generator-compose-v6-top-v3sub/generators/app/index.js
```diff
@@ -0,0 +1,21 @@
+"use strict";
+
+// A modern (yeoman-generator v6) top generator that composes a sub-generator
+// which only works on the legacy v3 runtime. Because the top is v6-based, the
+// whole run routes to v6 and the sub fails at writing time
+const Generator = require("yeoman-generator");
```
bd82: use aliased name

projects/yeoman-ui/packages/backend/test/fixtures/generator-compose-v6-top/generators/app/index.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+// A modern (yeoman-generator v6) top generator that composes a v6 sub-generator
+const Generator = require("yeoman-generator");
```
bd82: implict version

projects/yeoman-ui/packages/backend/test/utils/env.spec.ts
```diff
@@ -22,21 +23,45 @@ const envV6Fixture = {
   ),
 };
 
-const envV3Fixture = {
-  namespace: "env-v3-fixture:app",
-  packagePath: resolve(FIXTURES, "generator-env-v3-fixture"),
+const composeTopFixture = {
```
bd82: can you dynamically create the fixture tests?
particularly if we have distinct separation between positive/negative subfolders?

Probably not worth more than a little effort...

bd82: comment above is very optional

bd82 (changes requested):
just the logic comments and the fixtures name (for easy maintenance)
are the important bits

bd82 (dismissed):

rimasirich (approved):
LGTM

PR #583, opened 2026-08-24
chore: migrate vscode-logging

jacob-kreyenbuehl:
**What**
Consolidates the SAP/vscode-logging project (@vscode-logging/logger, /types, /wrapper + example packages) into app-studio-toolkit under projects/vscode-logging/, following the yeoman-ui (#503) and vscode-mta-tools (#547) precedents.

**Why**
Part of folding the separate BAS open-source repos into this monorepo: one CI, one security/Renovate setup, one Changesets release pipeline, less abandoned-repo drift.

**Behaviour**
- Packages at projects/vscode-logging/packages/*, examples at projects/vscode-logging/examples/*; pnpm-workspace.yaml + root package.json gain a projects/*/examples/* lookup. Examples are private (not published).
- Aligned to the monorepo toolchain (TypeScript 5.7.3, root prettier/eslint/husky/Changesets); the source repo's Lerna/husky/commitlint/prettier root config is dropped. A project-level README is kept.
- Emitting packages extend the root tsconfig.base.json. Instead of skipLibCheck, the root @types/node is bumped 16 -> 22 (oldest supported Node) to fix the @types/node-vs-DOM clashes at the source; this required a 2-line NodeJS.Timer -> NodeJS.Timeout fix in app-studio-toolkit's landscape.ts.
- The three packages are locked as a fixed Changesets group (mirrors the old Lerna locked mode); intra-repo deps use workspace:* everywhere. vscode-mta-tools' @vscode-logging/logger is switched to workspace:* so the Changesets dependents-graph stays consistent.
- Pre-migration changelogs renamed to CHANGELOG.old.md (project + per package, both listed in files); example changelogs removed; package READMEs point at the monorepo and drop the hand-written license section (legal-copy.js handles it).
- ESLint relaxed for projects/vscode-logging/** (same as yeoman-ui/mta-tools; TODO to tighten).
- scripts/merge-coverage.js now also merges projects/*/packages/* coverage.
- No functional change to the logger packages; all build and pass at 100% coverage. No changeset (release handled separately after merge).

Commits are staged move → format → fixes → examples for review; squash on merge.


rolanbadrislamov (dismissed):

projects/vscode-logging/examples/extension-wrapper/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
Cocossoul, line 1: To follow the pattern of other projects/packages, I'd assume we need to make this an extension of the tsconfig.base.json of the root

projects/vscode-logging/examples/extension/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
```
Cocossoul: This can be omitted, since the root gitignore already covers these cases

projects/vscode-logging/examples/extension/package.json
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+  },
+  "scripts": {},
+  "dependencies": {
+    "@vscode-logging/library-example": "workspace:*",
+    "@vscode-logging/logger": "workspace:^"
+  },
+  "devDependencies": {
+    "@types/vscode": "^1.32.0"
```
Cocossoul, line 57: Can we use the one from the root ? It is a newer version (1.75.0) so maybe not
If we really need a different version than the root, we could add a comment in the package.json to state why

The @types/vscode dev dependency is also added in every package.json of vscode-logging (not only this example)

projects/vscode-logging/examples/extension-wrapper/tsconfig.json
kubante: done.

projects/vscode-logging/examples/extension/package.json
kubante: about the version, this is on purpose. each package pins the minimum vscode api version it support (logger 1.51, wrapper 1.52), we keep it from the original repo. if we take the root one 1.75.0 the code could use newer apis and then break on older vscode, so we prefer not to. i can add a small note in the package.json to explain it if you think its more clear.

projects/vscode-logging/examples/extension/.gitignore
kubante: good catch, thanks. i removed it and also the same file in extension-wrapper (45b883f). the root gitignore already cover node_modules and *.vsix, and the examples dont build any vsix anyway.

projects/vscode-logging/CHANGELOG.old.md
```diff
@@ -0,0 +1,94 @@
+# Change Log
+
+> **Note:** This is the historic changelog from before vscode-logging was integrated into the app-studio-toolkit monorepo.
```
Cocossoul, line 3: Amazing comment ! To go further, how about we rename that file CHANGELOG.old.md ? This is the pattern that seems to be applied in the repo

jacob-kreyenbuehl: thanks! 

 in vscode-mta-tools the CHANGELOG.md is managed by changesets, because that project looks itself as a versioned package, so they parked the old pre migration history in CHANGELOG.old.md and let changesets own the live one. 

vscode-logging seams different, the folder itself is not a versioned package, changesets versions the packages inside packages/* and each one has its own changelog. so the changelog at the folder level looks like is only historic. 
yeoman-ui is the same kind of container and keeps it as a plain CHANGELOG.md with a note on top, so i followed that. I pinged Shachar to decide. if he prefers tge .old.md i d rename it right away.

projects/vscode-mta-tools/package.json
```diff
@@ -215,7 +215,7 @@
     "lodash": "4.17.21",
     "comment-json": "4.1.1",
     "fs-extra": "10.0.0",
-    "@vscode-logging/logger": "1.2.3",
+    "@vscode-logging/logger": "2.0.9",
```
bd82: I think this should be `workspace:*` syntax for intra-monorepo deps
Otherwise it would not update in the future.

This would be true for all consumers of "@vscode-logging/logger"
But we don't have to release all of them for e2e test, we could just release one of by limiting the changeset.

I would also consider changing the deps to: `workspace:*` in a separate PR after this one is merged.
because all the PR commits will be squashed into one commit on main branch.

projects/vscode-logging/CHANGELOG.old.md
jacob-kreyenbuehl: update: Shachar was ok with it, i renamed it to CHANGELOG.old.md . thanks both!

projects/vscode-logging/examples/extension-wrapper/tsconfig.json
jacob-kreyenbuehl: done. the extension-wrapper example now extend the root tsconfig.base.json and drop the duplicated options. i did the same for the wrapper package too.

projects/vscode-logging/examples/extension/.gitignore
jacob-kreyenbuehl: good catch, thanks. i removed it and also the same file in extension-wrapper (45b883f). the root gitignore already cover node_modules and *.vsix, and the examples dont build any vsix anyway.

projects/vscode-logging/examples/extension/package.json
jacob-kreyenbuehl: about the version, this is on purpose. each package pin the minimum vscode api version it support and we keep it from the original repo (this example 1.32, logger 1.51, wrapper 1.52). if we take the root one 1.75.0 the code could use newer apis and then break on older vscode, so we prefer to keep the lower pin. i can add a short note in the package.json to explain it if you think its more clear.

projects/vscode-mta-tools/package.json
jacob-kreyenbuehl: thanks, makes sense! i changed it to workspace:* now (69668d9), so mta-tools use the in-repo logger and it stay consistent. i checked and mta-tools still build and pass all its tests. for the other consumers that still use exact pins, i do the workspace:* conversion in a separate PR after this one is merged, like you suggested, so it doesnt mix into the squash here.

Cocossoul: I stacked a PR to do the same change in yeoman-ui https://github.com/SAP/app-studio-toolkit/pull/584

projects/vscode-logging/examples/extension/package.json
jacob-kreyenbuehl: done, i added a short note in the package.json (the "//@types/vscode" key) to explain why the version is pinned, like you suggested. i put it in the logger and wrapper packages and also the two extension examples (9dbece3). thanks!

bd82:
I think there is value to transfer (most of) the old root `README.md` even if its tiny.
For example see: https://github.com/SAP/app-studio-toolkit/tree/main/projects/vscode-mta-tools
Because with this `projects/*/pkgX` hierarchy it would explain what each project is.

For example: https://github.com/SAP/app-studio-toolkit/tree/main/projects/vscode-mta-tools

projects/vscode-mta-tools/package.json
```diff
@@ -215,7 +215,7 @@
     "lodash": "4.17.21",
     "comment-json": "4.1.1",
     "fs-extra": "10.0.0",
-    "@vscode-logging/logger": "1.2.3",
+    "@vscode-logging/logger": "workspace:*",
```
bd82, line 218: We might want a different *.vsix to test the release with.
Because VSCode-MTA-Tools was only maintained a little and bounced between teams.

But we need to change all references to `workspace:*` anyhow.
And testing the logging flow is easy on dev-space startup as there is normally a log on extension activation.

bd82: So, no change needed, just FYI

projects/vscode-logging/packages/wrapper/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../../../tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "dist",
+    "baseUrl": ".",
+    "skipLibCheck": true
```
bd82: the `skipLibCheck` should be avoided, its a heavy handed "disable button" for **all** external libs.
In this case its due to old "@types/node" dependency on the root of the report.
It seems resolved when updating to latest "@types/node" version **22**.

It makes sense to target oldest still supported nodejs version (v22 until April 27, v24 afterwards)

projects/vscode-logging/packages/wrapper/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+- `yarn add @vscode-logging/wrapper`
+
+## Usage
+
+Please see the [TypeScript Definitions](./api.d.ts) for full API details.
+
+## Example
+
+An example VSCode extension is available in the [examples' folder](https://github.com/SAP/vscode-logging/tree/master/examples/extension-wrapper).
```
bd82: link to old repo

projects/vscode-logging/packages/wrapper/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Support
+
+Please open [issues](https://github.com/SAP/vscode-logging/issues) on github.
```
bd82: link to old repo

projects/vscode-logging/packages/wrapper/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Contributing
+
+See [CONTRIBUTING.md](./CONTRIBUTING.md).
```
bd82: the link target is currently missing
should probably link to root contributing.md

projects/vscode-logging/packages/wrapper/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## License
```
bd82: the licensing and copyright notice is copied by: https://github.com/SAP/app-studio-toolkit/blob/main/scripts/legal-copy.js

This README.md `License` section is probably some legacy format, best to remove it to avoid duplication

projects/vscode-logging/packages/wrapper/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "bugs": {
+    "url": "https://github.com/SAP/app-studio-toolkit/issues"
+  },
+  "license": "Apache-2.0",
+  "typings": "./api.d.ts",
+  "dependencies": {
+    "@vscode-logging/logger": "workspace:^",
+    "@vscode-logging/types": "workspace:^"
+  },
+  "//@types/vscode": "Pinned to the minimum VS Code API version supported here (kept from the standalone vscode-logging repo). Intentionally not raised to the monorepo root version, which would allow newer APIs and break support for older VS Code.",
```
bd82, line 32: This `pinning` and comment is technically correct.

But it raises a centralization management concern for all these `@types/vscode` devDeps.
Random thoughts:

1. Should we pin for each new project we import? and manage multiple different `pins
2. In practice VSCode published APIs rarely break as opposed to ["Proposed APIs"](https://code.visualstudio.com/api/advanced-topics/using-proposed-api), so we might be "protecting against a non-issue"
3. Would it be simpler to maintain a single `@types/vscode` at the root of this monorepo (one already exists).
4. These specific packages (vscode-logging) are used inside other VSCode extensions, so they might be the exception to the rule as "libraries". But other *.vsix packages in this monorepo primirialy need to work in BAS / Modern VSCode  (self contained) so a-lot less need to `pin` to (very) old versions.

**No action is needed right now. its not part of the scope of this PR.**
Just food for thought on this **growing** mono-repo management.

.eslintrc.js
```diff
@@ -178,6 +189,19 @@ module.exports = {
         "@typescript-eslint/no-misused-promises": "off",
       },
     },
+    {
+      // Additional TypeScript rules for vscode-logging packages.
+      // TODO: clean up violations and tighten incrementally.
+      files: ["projects/vscode-logging/**/*.ts"],
+      rules: {
+        // public API surface (api.d.ts) intentionally exposes `any`
+        "@typescript-eslint/no-explicit-any": "off",
```
bd82, line 198: We are disabling some eslint rules because of more strict scanning in this monorepo.
So we still gain improved static checks while incrementally aligning to the central mono-repo standard.

In this case if we already discovered that is causing the issue and it is intentional.
Instead of disable + documenting it in a comment
we can just resolve it via a comment at the top of each of those `api.d.ts` files
- `/* eslint-disable @typescript-eslint/no-explicit-any */`

This prevents disabling this very useful eslint rule for the entire vscode-logging project.

Its recommended but not mandatory... Can be deferred to later eslint improvement pass.

projects/vscode-logging/packages/wrapper/CHANGELOG.old.md
```diff
@@ -0,0 +1,88 @@
+# Change Log
+
```
bd82, line 2: I am not sure the changeSets generated changelog will integrate with the previous conventional commits changelog.

I believe it caused issues in previous migrations and so we renamed the old file to CHANGELOG.old.md
See in: https://github.com/SAP/app-studio-toolkit/tree/main/projects/yeoman-ui/packages/backend

bd82: Note as this package is published to npmjs.com
You will need to update the `files` section in the package.json
- https://docs.npmjs.com/cli/v11/configuring-npm/package-json#files
so both the new and old changelogs would be included in the published artifact

projects/vscode-logging/packages/types/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "noEmit": true,
+    "target": "ES2015",
+    "skipLibCheck": true
```
bd82: See: https://github.com/SAP/app-studio-toolkit/pull/583/changes#r3855304237

In this case I also had to remove `    "target": "ES2015",` to get it to compile
this target is redundant anyhow because noEmit is enabled. so there is no actual *.js output

projects/vscode-logging/packages/types/README.md
```diff
@@ -0,0 +1,49 @@
+# @vscode-logging/types
+
```
bd82, line 2: See comments on similar README.md file here:
- https://github.com/SAP/app-studio-toolkit/pull/583/changes#r3855304237

projects/vscode-logging/packages/types/CHANGELOG.old.md
```diff
@@ -0,0 +1,70 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
```
bd82, line 3: see: https://github.com/SAP/app-studio-toolkit/pull/583/changes#r3855617051

projects/vscode-logging/packages/logger/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "noEmit": true,
+    "checkJs": true,
+    "allowJs": true,
+    "skipLibCheck": true
```
bd82: see: https://github.com/SAP/app-studio-toolkit/pull/583/changes#r3855304237

projects/vscode-logging/packages/logger/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+- Processes which manipulate the source code such as [bundling][vscode-bundling] or compilation (TypeScript/Babel)
+  may make the information produced less relevant due to the lack of sourceMaps support.
+  Therefore this feature may be more useful during development flows rather then productive flows.
+
+[vscode-bundling]: https://code.visualstudio.com/api/working-with-extensions/bundling-extension
+
+## Support
+
+Please open [issues](https://github.com/SAP/vscode-logging/issues) on github.
```
bd82: link to old repo

projects/vscode-logging/packages/logger/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Contributing
```
bd82, line 139: see comments on similar README.md
[README.md](https://github.com/SAP/app-studio-toolkit/pull/583/changes#diff-4291bab4553e346d112d1a4f53c5abe506d3d3e0b010df521a0616c5d7acaa12)

projects/vscode-logging/examples/extension-wrapper/CHANGELOG.md
```diff
@@ -0,0 +1,86 @@
+# Change Log
```
bd82: I don't think the changelog matters for the non published examples.
And will likely cause issues with changeSets CHANGELOG
so deletion is probably safest

projects/vscode-logging/examples/extension/CHANGELOG.md
```diff
@@ -0,0 +1,111 @@
+# Change Log
```
bd82: I don't think the changelog matters for the non published examples.
And will likely cause issues with changeSets CHANGELOG
so deletion is probably safest

projects/vscode-logging/examples/library/CHANGELOG.md
```diff
@@ -0,0 +1,67 @@
+# Change Log
```
bd82: I don't think the changelog matters for the non published examples.
And will likely cause issues with changeSets CHANGELOG
so deletion is probably safest

projects/vscode-logging/packages/logger/CHANGELOG.old.md
```diff
@@ -0,0 +1,125 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
```
bd82, line 3: See: https://github.com/SAP/app-studio-toolkit/pull/583/changes#r3855617051

projects/vscode-logging/examples/library/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@vscode-logging/library-example",
+  "version": "2.0.8",
+  "private": true,
+  "description": "Example node library using @vscode-logging/logger via Dependency Injection",
+  "main": "index.js",
+  "repository": "https://github.com/sap/vscode-logging/",
+  "license": "Apache-2.0",
+  "devDependencies": {
+    "@vscode-logging/types": "workspace:^"
```
bd82: I don't think any issue will be caused because this is not published but why should we use a range `workspace:^`
instead of a fixed pinned `workspace:*`?

Probably better to use `workspace:*` assuming we intend to keep the three vscode-logging productive packages "locked" with shared release life-cycle.

Not because this will cause any issue, but because of the **edge case** (arguably extreme...) somebody will copy/paste
this syntax into a productive package and cause unexpected runtime dependencies resolution.
- Assuming all `workspace:X` that will ever exist in this project will be of shape: `workspace:*`

projects/vscode-logging/examples/extension/package.json
```diff
@@ -0,0 +1,59 @@
...
+          "description": "Should Source Code Location Info be added to log entries, DANGER - May be very slow, only use in debugging scenarios",
+          "scope": "resource"
+        }
+      }
+    }
+  },
+  "scripts": {},
+  "dependencies": {
+    "@vscode-logging/library-example": "workspace:*",
+    "@vscode-logging/logger": "workspace:^"
```
bd82: why is one workspace: `^` and the other `*`?
I think pinned (`*`) is best for us (locked shared version for the 3-6 vs-code packages).

projects/vscode-logging/examples/extension-wrapper/package.json
```diff
@@ -0,0 +1,63 @@
...
+      }
+    }
+  },
+  "dependencies": {
+    "@vscode-logging/wrapper": "workspace:^"
+  },
+  "//@types/vscode": "Pinned to the minimum VS Code API version supported here (kept from the standalone vscode-logging repo). Intentionally not raised to the monorepo root version, which would allow newer APIs and break support for older VS Code.",
+  "devDependencies": {
+    "@types/vscode": "^1.52.0",
+    "@vscode-logging/types": "workspace:^"
```
bd82: workspace:* ?

scripts/merge-coverage.js
```diff
@@ -3,7 +3,7 @@
  * This report will can be found at the root `coverage` dir and will also be uploaded to coveralls.io
  * - based on https://github.com/istanbuljs/istanbuljs/blob/1fe490e51909607137ded25b1688581c9fd926cd/monorepo-merge-reports.js
  */
```
bd82, line 5: Random thought: No action needed.

In a larger monorepo with many projects/packages is a merged coverage report still needed?
- Particularly with many sub-projects being legacy? 
- not all imported projects will meet the 100% tresh-hold anyhow, so should each project control its own coverage treshold gate?

bd82 (changes requested):
See inline comments.

I also think the changesets config should be modified to lock the new packages as a single release, so if one is bumped all are bumped.
See: https://github.com/SAP/app-studio-toolkit/blob/main/.changeset/config.json#L8-L13

The reason is because the previous vscode-logging lerna.json was set to the default["fixed/locked"](https://lerna.js.org/docs/features/version-and-publish#fixedlocked-mode-default) and this would be the most similar flow with changeSets.

In theory unlocked can result in strange edge cases like multiple versions of the same package existing at the same time at runtime.

I'm not sure if the examples should be part of this lock.
1. Their version means little, they are not published.
2. Would locking them create more github tags spam when one of the three productive packages are released

projects/vscode-logging/packages/wrapper/tsconfig.json
jacob-kreyenbuehl: good point, done. i bumped the root @types/node to 22.20.1 and removed skipLibCheck from logger, types and wrapper (f552c2e). you were right, it was only the old @types/node. one ripple: v22 changes the setInterval return type, so app-studio-toolkit landscape.ts needed 2 small annotations (NodeJS.Timer -> NodeJS.Timeout). i included them in the same commit since the bump forces it. full ci is green.

projects/vscode-logging/packages/types/tsconfig.json
jacob-kreyenbuehl: done (f552c2e). removed skipLibCheck and also the redundant target ES2015 like you said, it compiles fine without it since noEmit is on.

projects/vscode-logging/packages/logger/tsconfig.json
jacob-kreyenbuehl: done here too (f552c2e). see the wrapper tsconfig thread for the @types/node bump details.

projects/vscode-logging/packages/wrapper/README.md
jacob-kreyenbuehl: done (de9c186), points to the monorepo now.

projects/vscode-logging/packages/wrapper/README.md
jacob-kreyenbuehl: done (de9c186), points to the monorepo now.

projects/vscode-logging/packages/wrapper/README.md
jacob-kreyenbuehl: done (de9c186), links to the root CONTRIBUTING.md now.

projects/vscode-logging/packages/wrapper/README.md
jacob-kreyenbuehl: done (de9c186), removed the license section since scripts/legal-copy.js injects it at build.

projects/vscode-logging/packages/types/README.md
jacob-kreyenbuehl: done (de9c186), same fixes as the wrapper readme (monorepo links, contributing to root, removed the license section).

projects/vscode-logging/packages/logger/README.md
jacob-kreyenbuehl: done (de9c186), points to the monorepo now.

projects/vscode-logging/packages/logger/README.md
jacob-kreyenbuehl: done (de9c186), same fixes as the wrapper readme (monorepo links, contributing to root, removed the license section).

projects/vscode-logging/packages/wrapper/CHANGELOG.old.md
jacob-kreyenbuehl: done, renamed the 3 package changelogs to CHANGELOG.old.md so changesets can own the fresh one (b8e2f53), and added both CHANGELOG.md and CHANGELOG.old.md to the files field so they still ship to npm.

projects/vscode-logging/packages/types/CHANGELOG.old.md
jacob-kreyenbuehl: done (b8e2f53), same as the wrapper changelog thread.

projects/vscode-logging/packages/logger/CHANGELOG.old.md
jacob-kreyenbuehl: done (b8e2f53), same as the wrapper changelog thread.

projects/vscode-logging/examples/extension-wrapper/CHANGELOG.md
jacob-kreyenbuehl: agreed, deleted it (b8e2f53).

projects/vscode-logging/examples/extension/CHANGELOG.md
jacob-kreyenbuehl: agreed, deleted it (b8e2f53).

projects/vscode-logging/examples/library/CHANGELOG.md
jacob-kreyenbuehl: agreed, deleted it (b8e2f53).

projects/vscode-logging/examples/library/package.json
jacob-kreyenbuehl: you re right, changed it to workspace:* (c8c2df5). we made all of them * now for consistency, examples included.

projects/vscode-logging/examples/extension/package.json
jacob-kreyenbuehl: fixed, both are workspace:* now (c8c2df5).

projects/vscode-logging/examples/extension-wrapper/package.json
jacob-kreyenbuehl: done, workspace:* now (c8c2df5).

.eslintrc.js
jacob-kreyenbuehl: makes sense. i kept the project override for now, but noted to move no-explicit-any to a per-file /* eslint-disable */ on the api.d.ts files in a later eslint pass, like you said (recommended not mandatory).

projects/vscode-logging/packages/wrapper/package.json
jacob-kreyenbuehl: thanks, good points. no change now as you said. we keep the pin plus the //@types/vscode note for these library packages, and the central @types/vscode question is worth a repo-wide discussion later.

scripts/merge-coverage.js
jacob-kreyenbuehl: agree, no action now. worth revisiting how the coverage gate should work per project as the monorepo grows.

jacob-kreyenbuehl:
thanks for the thorough review @bd82, really helpful. i pushed 6 commits addressing everything:

- **@types/node -> 22 + removed skipLibCheck** (f552c2e). you were right it was only the old @types/node. the bump also needed 2 small type fixes in app-studio-toolkit landscape.ts (Timer -> Timeout), included in the same commit.
- **fixed changeset group** for logger/types/wrapper, examples excluded (056ad90). note: types is on 2.0.8 vs 2.0.9 for the other two, the fixed group will line them up on the first release.
- **per-package changelogs -> CHANGELOG.old.md** + added both to the files field, and deleted the example changelogs (b8e2f53).
- **workspace:* everywhere** (c8c2df5).
- **README fixes** on the 3 packages (de9c186) and a **new project README** (b99a080).

deferred as you suggested: the per-file eslint-disable on api.d.ts (later eslint pass). the @types/vscode centralization and coverage-gate questions i left as no-action, food for thought for later.

full ci is green and the branch is up to date with main. i left your threads open so you can verify. ready for another look when you have time.

bd82 (approved):

bd82 (approved):

PR #591, opened 2026-08-26
fix(yeomen-ui): bump inquirer-gui to 3.4.14 to fix text clipping in input fields

Artemis-Coder17:
## Changes
- Bumped `@sap-devx/inquirer-gui` and related plugins from `3.4.13` to `3.4.14`

### Changes in `Inquirer-Gui`
- Added `applyPatch` function in `QuestionInput.vue` that injects a `<style>` tag into the `vscode-textfield` shadow root on `mounted()`
- The injected style overrides the hardcoded `line-height: 18px` on the inner `<input>` with `line-height: normal`, allowing the browser to calculate the natural line height for the active font

## Problem
Text in `<vscode-textfield>` inputs was being clipped at the bottom on Windows (Segoe UI font) / Characters with descenders (`g`, `y`, `p`, `q`, `j`) had their tails cut off.

**Root cause:** `@vscode-elements/elements@1.11.0` hardcodes `line-height: 18px` on the inner `<input>`. At `16px` font size (set by `inquirer-gui`), Segoe UI descenders exceed this height and get clipped.


alex-gilin (approved):
LGTM

PR #596, opened 2026-08-28
fix: mdk app generator issue

ilya-taupeka:


projects/yeoman-ui/packages/yeoman-env-v3/webpack.config.cjs
```diff
@@ -86,6 +86,20 @@ module.exports = {
...
+      // dynamic imports in a CommonJS bundle into a module lookup, which breaks
+      // runtime loading of generator .mjs files from disk ("Cannot find module
+      // 'file://...'"). Hide import() from webpack so Node handles it natively.
+      {
+        test: /yeoman-environment[/\\]lib[/\\]util[/\\]esm\.js$/,
+        loader: "string-replace-loader",
+        options: {
+          search: "return import\\(pathToFileURL\\(fileToImport\\)\\);",
+          replace:
+            "return new Function('specifier', 'return import(specifier)')(pathToFileURL(fileToImport).href);",
```
bd82, line 99: instead of wrapping with "new Function" which can fail in runtimes with security policies.
Can we use: webpack ignore comment?

- https://webpack.js.org/api/module-methods#magic-comments

bd82: https://chatgpt.com/s/t_6a93da5742e48191ad50b7de8455ce30

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -89,14 +90,30 @@ class EnvUtil {
   //    write a "resolved" property onto the generator module's frozen ESM
   //    namespace object, which throws. This is an ESM/CJS incompatibility, not
   //    a version guard.
+  //
+  // 3. "Cannot find module 'file://.../*.mjs'" - the bundled v3 compatibility
```
bd82, line 94: if the v3 bundling was already resolved in webpack.config
why do we need to add this "Cannot find module" handling 
to the already none trivial logic in the env selection.

And assuming there is a valid need to add this error handling.
How can you guarantee this totally generic error: "Cannot find module" would not
later be needed to identify V6 or (vN) env versions.

This logic seems brittle

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -296,31 +321,49 @@ class EnvUtil {
...
+      }
+    );
+    v3Env.register(resolved, {
       namespace: genNamespace,
       packagePath: meta.packagePath,
     } as any);
-    const gen = v3Env.create(genNamespace, {
-      options,
-    } as unknown as string[]);
+    const gen = await Promise.resolve(
```
bd82, line 362: Previously this was not wrapped in a promise.
Why the change? I don't object, Just trying to understand the scope of the change.

projects/yeoman-ui/packages/backend/test/utils/env.spec.ts
```diff
@@ -219,6 +247,79 @@ describe("Env.createRunGen()", () => {
...
+    expect(prepare.firstCall.args[1]).to.deep.equal({ id: "v6" });
+  });
+
+  it("runs on v6 when bundled v3 cannot load an ESM generator file URL", async () => {
+    sandbox
+      .stub(Env as any, "createLegacyV3EnvAndGen")
+      .rejects(
+        Object.assign(
+          new Error(
+            "Cannot find module 'file:///extbin/globals/bun/global/install/node_modules/@ext-mdkvsc-npm-rel/generator-mdk/generators/app/index.mjs'"
```
bd82, line 278: after fixing the webpack.config can this really happen in production without stubs/mocks?

projects/yeoman-ui/packages/backend/src/utils/env.ts
```diff
@@ -89,14 +90,30 @@ class EnvUtil {
...
+  //    rewrites v3's native dynamic import into a CommonJS module lookup. Treat
+  //    only file-URL ESM module failures as a v3-runtime signal; ordinary missing
+  //    package errors must still surface as real generator errors.
   private isV3RuntimeIncompatibilityError(error: unknown): boolean {
     const message = (error as Error)?.message ?? "";
+    const isFrozenEsmNamespaceError =
+      message.includes("object is not extensible") &&
+      message.includes("resolved");
+    const isBundledV3EsmImportError =
+      /Cannot find module ['"]file:\/\/.*\.(mjs|mts)['"]/.test(message);
```
bd82, line 105: This condition raises questions:
- `/Cannot find module ['"]file:\/\/.*\.(mjs|mts)['"]/.test(message)`

1. Why only testing for `mjs/mts` files? esm generators can also appear as *.js files.
2. Could this solution be too specific for a the mdk generator error and not generic?

projects/yeoman-ui/packages/backend/test/utils/env.spec.ts
```diff
@@ -219,6 +247,79 @@ describe("Env.createRunGen()", () => {
...
+      { silent: true },
+      fakeAdapter(),
+      prepare
+    );
+
+    expect(v6Create.calledOnce, "the generator fell back to v6").to.equal(true);
+    expect(prepare.firstCall.args[1]).to.deep.equal({ id: "v6" });
+  });
+
+  it("runs on v6 when the bundled v3 ESM load error arrives wrapped", async () => {
```
bd82, line 299: can you explain where is the "wrapping" here?
what does it add beyond the previous test (which uses "object.assign")

bd82 (changes requested):
Main things:

1. It seems `env.ts` has logic to handle edge case that was resolved via webpack.config modification.
2. It seems the logic is too **specific** to `*.mjs` files, but ESM is not limited to `*.mjs` files.
3. It seems the logic is too **generic** due to handling `cannot find module` error, but in this case how can we guarantee such logic won't conflict with other env versions in the future, or to even distinguish with a real `cannot find moudle` from V6.

bd82:
replaced with: https://github.com/SAP/app-studio-toolkit/pull/599

PR #601, opened 2026-09-01
chore: migrate guided-development

rolanbadrislamov:
## Summary

- Moves `SAP/guided-development` into `projects/guided-development/`.
- Splits into three packages: `packages/backend` (publishable VSCode extension), `packages/types` (publishable `@sap_oss/guided-development-types`), `packages/frontend` (private Vue 3 app bundled into backend)
- Six example/demo extensions land under `examples/` as `private: true`

## Key changes

- All sibling deps use `workspace:*` protocol (no exact-version pins that would break `changeset version`)
- Root `.eslintrc.js` extended with `projects/guided-development/**` overrides to handle legacy TSLint-era code without rewriting source logic
- Replaced broken `import { bas }` (ambient-only, not exported) with `import { BasToolkit }` throughout backend and tests
- Removed `skipLibCheck: true` from all tsconfigs; added `@types/minimatch@^3.0.5` to example packages to satisfy `@types/glob@7` transitive dep
- Deleted per-package `.eslintrc` (had `"root": true` which blocked the root config from applying)


.gitignore
```diff
@@ -77,5 +77,9 @@ tsconfig.tsbuildinfo
 # generated artifacts
 dist
 
+# guided-development: tsc `out/` output and nyc `reports/` coverage
+projects/guided-development/**/out
```
bd82, line 81: you can have a nested `.gitignore` ifle

.eslintrc.js
```diff
@@ -202,6 +223,35 @@ module.exports = {
         "@typescript-eslint/no-unsafe-argument": "off",
       },
     },
+    {
+      // Additional TypeScript rules for guided-development packages.
+      // Legacy code (formerly TSLint / ts@3.9) migrated as-is; relaxed to match
+      // the historical behaviour without rewriting source logic.
+      // TODO: clean up violations and tighten incrementally.
+      files: ["projects/guided-development/**/*.ts"],
```
bd82, line 231: This eslint configuration is getting a bit large and difficult to manage.
Maybe split the new configurations into separate file (per project?)

bd82: Also true and more important that if the consolidation work is **parallelized** 
then we need the multiple files (eslint-config/guided-dev.js) approach to reduce merge conflicts.

.changeset/config.json
```diff
@@ -28,7 +29,15 @@
     "yeoman-env-v3",
     "vscode-logging-extension-example",
     "vscode-logging-extension-wrapper-example",
-    "@vscode-logging/library-example"
+    "@vscode-logging/library-example",
+    "@sap-devx/guided-development-root",
+    "guided-development-frontend",
```
bd82, line 34: frontend package is not an example, its a real production package that is bundled inside the backend.
It's out of scope to decide if we want to only manage CHANGELOGs for the *.vsix packages.
Might be a good thing, but no capacity / scope to explore this now.

projects/guided-development/package.json
```diff
@@ -0,0 +1,4 @@
+{
```
bd82, line 1: this seems redundant, as this root package.json is empty
we can add it later if needed

projects/guided-development/README.md
```diff
@@ -0,0 +1,36 @@
+[![Continuous Integration](https://github.com/SAP/app-studio-toolkit/actions/workflows/ci.yml/badge.svg)](https://github.com/SAP/app-studio-toolkit/actions/workflows/ci.yml)
```
bd82, line 1: Open Question / Alignment:

- Do we need all these badges?
- Do we want to align all the bandages between all the projects? (what about in the future)?
- Would it be simpler to just remove the badages as they should already exist in the repo root README.md?

projects/guided-development/packages/types/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@sap_oss/guided-development-types",
```
bd82, line 2: this package seems to be deprecated

Image

- So, do we need a new scope to release it?
- Who is using it (internally / externally)?
- Should we even do a new release as part of consolidation?
- If we do new release, do consumers need to update their deps?
- What is the functionality? is still still in PROD?

bd82: these are only design time types. it might be a valid decision to stop releasing them
or it might not be a big impact to release in a new scope.

Note that releasing in a new scope will require inital manual release before trusted publishers can be configured in the npmjs UI.

projects/guided-development/packages/backend/package.json
```diff
@@ -0,0 +1,181 @@
+{
+  "name": "guided-development",
```
bd82, line 2: this is a *.vsix, it should be marked as `private`
otherwise it might be release to npmjs.com which is not the intent

projects/guided-development/packages/backend/package.json
```diff
@@ -0,0 +1,181 @@
...
+    "frontend:build": "pnpm --filter guided-development-frontend run build",
+    "frontend:copy": "shx mkdir -p ./dist/media && shx cp -r ../frontend/dist/. ./dist/media/",
+    "package": "vsce package --no-dependencies",
+    "ws:run": "node ./out/src/webSocketServer/index.js"
+  },
+  "dependencies": {
+    "@sap-devx/app-studio-toolkit-types": "workspace:*",
+    "@sap-devx/webview-rpc": "0.3.1",
+    "@sap_oss/guided-development-types": "workspace:*",
+    "@vscode-logging/logger": "workspace:*",
```
bd82, line 155: previously we did using "workspace:*" for vscode-logging
because this will cause cascading large releases of everything everytime vscode-logging changes
instead we are keeping the manual bumping.

So the version should point to the one on npmjs.com

projects/guided-development/packages/backend/package.json
```diff
@@ -0,0 +1,181 @@
...
+    "chalk": "^4.0.0",
+    "datauri": "^3.0.0",
+    "fs-extra": "10.0.0",
+    "humanize-string": "^2.1.0",
+    "lodash": "4.17.21",
```
bd82, line 160: this version has a security bug (lodash)

projects/guided-development/packages/backend/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "extends": "../../tsconfig.base.json",
```
bd82, line 2: this looking like a bit hybird approach
1. We are exetnding the base tsconfig.json
2. But we still have a very custom project level tsconfig.json

Maybe its better to keep the original typescript version/config in this case
without linking it to the root project settings
and later solve the tech debt of single aligned typescript version / config?

projects/guided-development/packages/backend/README.md
```diff
@@ -0,0 +1,22 @@
+# Extension
+
+## Project setup
+
+```
+npm install
```
bd82, line 6: remove the project setup instructions

projects/guided-development/packages/backend/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Launch the WebSocket server configuration for development purposes. This will not launch vscode or a vscode extension, but still communicate with a locally served Vue application.
+
+Launch the _Run Dev Server_ launch configuration.
+
+Or use the command line:
+
+```sh
+# compile server code to out directory:
+npm run compile
```
bd82, line 19: these should probably be pnpm commands now

projects/guided-development/packages/frontend/package.json
```diff
@@ -0,0 +1,104 @@
...
+    "@sap-devx/inquirer-gui-login-plugin": "3.0.0",
+    "@sap-devx/inquirer-gui-tiles-plugin": "3.0.0",
+    "core-js": "^3.18.3",
+    "material-design-icons-iconfont": "^5.0.1",
+    "vue": "^3.3.4",
+    "vue-loading-overlay": "^6.0.3",
+    "vuetify": "^3.3.10"
+  },
+  "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^6.2.0",
```
bd82, line 42: why are there eslint plugins in dev-deps?
is not this handled by the root package.json?

projects/guided-development/packages/frontend/package.json
```diff
@@ -0,0 +1,104 @@
...
+    "jest-serializer-vue": "^2.0.2",
+    "mock-socket": "^9.0.2",
+    "sass": "^1.26.3",
+    "sass-loader": "^8.0.2",
+    "ts-jest": "^27.0.4",
+    "vue-template-compiler": "^2.6.10",
+    "webpack-plugin-vuetify": "^2.0.0",
+    "webpack": "^5.0.0"
+  },
+  "eslintConfig": {
```
bd82, line 67: also eslint config, looks like it should be extract to root config (or alreayd was)?

projects/guided-development/packages/frontend/README.md
```diff
@@ -0,0 +1,35 @@
+# Guided Development Frontend
```
bd82, line 1: pnpm commands, try to see if 1-2 actually work

bd82 (dismissed):
see inline comments

bd82:
So the `sap_oss` scope will not become a blocker
it is possible to just mark the types package as `private` and even exlcude it from changesets config.
and leave it as tech debt to be resolved later.

.eslintrc.js
rolanbadrislamov: Agreed, will do the split as a follow-up PR that refactors all existing overrides (yeoman-ui, vscode-logging, guided-development) into per-project files at once to keep it consistent.

projects/guided-development/packages/frontend/package.json
rolanbadrislamov: The frontend uses Vue CLI's own ESLint pipeline which requires these plugins specifically (@vue/cli-plugin-eslint, babel-eslint, eslint-plugin-vue). The root ESLint config doesn't cover Vue CLI's build-time linting. These need to stay here

projects/guided-development/packages/backend/package.json
rolanbadrislamov: Fixed, added `private: true`.

projects/guided-development/packages/backend/tsconfig.json
rolanbadrislamov: Reverting to the original would require bringing back `skipLibCheck: true` which was flagged in a previous review. The hybrid is intentional for now. I will clean up redundant options in a follow-up

projects/guided-development/packages/backend/package.json
```diff
@@ -0,0 +1,182 @@
...
+    "bundle": "webpack --mode production",
+    "webpack": "webpack --mode development",
+    "webpack-dev:watch": "webpack --mode development --watch",
+    "frontend:build": "pnpm --filter guided-development-frontend run build",
+    "frontend:copy": "shx mkdir -p ./dist/media && shx cp -r ../frontend/dist/. ./dist/media/",
+    "package": "vsce package --no-dependencies",
+    "ws:run": "node ./out/src/webSocketServer/index.js"
+  },
+  "dependencies": {
+    "@sap-devx/app-studio-toolkit-types": "workspace:*",
```
jacob-kreyenbuehl: this is the one worth deciding on. `app-studio-toolkit-types` lives in another project, and our vscode-logging rule was "workspace:* inside a project, pin across projects", so by that it'd be pinned `"2.11.4"`. `app-studio-toolkit` and `vscode-deps-upgrade-tool` already use `workspace:*` for it though, so this is consistent either way.

catch: `bumpVersionsWithWorkspaceProtocolOnly` only shields pinned deps from the cascade. so with `workspace:*`, every time `app-studio-toolkit-types` releases, guided-development gets pulled into that release wave too. pin it if we want this project to release on its own, leave it if we're fine riding along. its private/vsix only so this is just version churn. i'd lean pin, but no strong opinion.

.changeset/config.json
```diff
@@ -28,7 +29,14 @@
     "yeoman-env-v3",
     "vscode-logging-extension-example",
     "vscode-logging-extension-wrapper-example",
-    "@vscode-logging/library-example"
+    "@vscode-logging/library-example",
+    "@sap-devx/guided-development-root",
```
jacob-kreyenbuehl: this entry matches nothing, the root package.json was correctly not materialized so `@sap-devx/guided-development-root` isnt a package in the workspace. safe to drop this one line (keep the frontend + example entries).

projects/guided-development/packages/backend/package.json
rolanbadrislamov: Pinned to `2.11.4` to keep releases independent

.changeset/config.json
rolanbadrislamov: Removed, good catch

projects/guided-development/packages/types/package.json
jacob-kreyenbuehl: @rolanbadrislamov this thread is the only one still open i think, and its the one that could block a release, so worth closing out while shachar is away.

shachar already gave the escape hatch in a comment above: mark the types package `private: true` and drop it from the changeset config (out of `fixed`, into `ignore`), leave it as tech debt. that avoids the whole "do we need the `@sap_oss` scope / should we even re-release deprecated design-time types" question for now, and it unblocks the migration without making a call thats really his to make.

my vote: do exactly that (private + ignore) for this PR, and open a small follow-up issue to decide the real fate of the package (keep releasing / deprecate / new scope) when hes back. that keeps guided-development releasing cleanly (backend vsix + nothing else) and doesnt commit us to anything.

if the types are consumed externally though, private is a breaking change for those consumers, so worth a quick check first: do we know who pulls `@sap_oss/guided-development-types` (or the old `@sap-devx` one)? if nobody external, private is a no-brainer.

rolanbadrislamov: Done, marked `private: true` and moved to ignore in the last commit. Checked npm: the package is already deprecated with only one version ever published. We can decide the fate of the types later

jacob-kreyenbuehl (dismissed):
All review comments addressed. Approving per team self-merge authority.

jacob-kreyenbuehl (approved):
All review comments addressed, changeset config fixed (backend moved to fixed group). Ready to merge.

jacob-kreyenbuehl (dismissed):

projects/guided-development/packages/backend/.gitignore
```diff
@@ -0,0 +1,5 @@
+dummy
+LICENSES
```
Cocossoul: .eslintcache and LICENSES are already covered by the root gitignore

projects/guided-development/packages/types/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
```
Cocossoul: Can be deleted, already covered by root gitignore

projects/guided-development/packages/backend/.gitignore
jacob-kreyenbuehl: Fixed in 3f23213 - deleted the file. Both .gitignore files were fully redundant (root covers LICENSES, .eslintcache, node_modules, dist).

projects/guided-development/packages/types/.gitignore
jacob-kreyenbuehl: Fixed in 3f23213 - deleted the file.

projects/guided-development/packages/backend/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Running during development
+
+Launch the _Run Dev Server_ launch configuration.
+
+Or use the command line:
+
+```sh
+# compile server code to out directory:
+pnpm run compile
```
Cocossoul, line 13: Compilation fails until the types are "compiled" too

projects/guided-development/packages/types/package.json
```diff
@@ -0,0 +1,52 @@
...
+    ".reuse",
+    "LICENSES",
+    "dist",
+    "CHANGELOG.md",
+    "CHANGELOG.old.md"
+  ],
+  "scripts": {
+    "ci": "npm-run-all clean compile",
+    "clean": "shx rm -rf ./dist",
+    "compile": "npm-run-all compile:types compile:esm compile:cjs",
```
Cocossoul, line 37: I'm not sure what's the intent here, I'm a bit confused about compiling types: types should only be a declaration, no compilation needed.
This wasn't present in the original repository, any reason why it was added here ?

projects/guided-development/.gitignore
```diff
@@ -0,0 +1,5 @@
+# tsc output
+**/out
```
Cocossoul, line 2: It could be worth standardize tsc output and nyc coverage reports output directories between all the packages, in a subsequent PR of course

rolanbadrislamov: Agreed, standardizing output directories (out vs dist) across all packages would be a good cleanup. Out of scope for this migration PR, but happy to do it as a follow-up

projects/guided-development/packages/types/package.json
rolanbadrislamov: The types package here is not purely ambient declarations, it has a real runtime source `(src/index.ts)` that exports constants and interfaces contributors need at runtime. In the original repo this source lived in `backend/src/types.ts` and was published to npm via an `ncp copy hack`. During migration it was extracted into its own proper package, so it now needs a real build (esbuild for ESM/CJS + tsc for .d.ts). The compilation is intentional

projects/guided-development/packages/backend/README.md
rolanbadrislamov: Fixed, updated the README to clarify that `packages/types` must be compiled first before the backend, since it's a workspace dependency. Added the explicit `pnpm --filter` command to make this clear.

Cocossoul (approved):

PR #619, opened 2026-09-17
chore: migrate task-explorer

deekshas8:
### What

Migrates SAP/task-explorer into this monorepo under projects/task-explorer/, bringing 5 packages:

  - Published (1): `@sap_oss/task_contrib_types `
  - Private/support (4): `vscode-tasks-explorer-tasks-panel` (the extension), `@vscode-tasks-explorer/vue_frontend_rpc` (the Vue 3 webview frontend) and the two example extensions `npm-task-contrib` and `vscode-task-contrib`.
 
###  Why

  Consolidates the task-explorer tooling into app-studio-toolkit alongside its consumers, so it shares the monorepo's pnpm/TypeScript toolchain, CI, and Changesets release pipeline instead of living in a separate yarn-based repo.

### Behaviour

  - Pure relocation, no functional changes to the packages.
  - Wired into the pnpm workspace: intra-project deps use workspace:* (tasks_panel → task_contrib_types + vue_frontend_rpc; both examples → task_contrib_types).
  - pnpm run ci green end-to-end: format + lint clean, all package test suites pass, VSIX packaging succeeds, merged coverage at 100%.
  - No changeset in this PR (pure relocation; release is a separate post-merge step).

### Reviewer's guide (192 files, but only ~10 matter)

  Most of the diff is a straight source lift (src/, test/, Vue components, icons, READMEs/CHANGELOGs). The signal is in the integration points. Suggested order:
  1. Release wiring (review thoroughly — this is what can break main)
  - .changeset/config.json — the 3 private packages (@vscode-tasks-explorer/vue_frontend_rpc, npm-task-contrib, vscode-task-contrib) added to ignore; the 2 publishable packages are not in a fixed group (removed in 0c7d448 — they version independently). No changeset .md (release is a separate step).
  
  2. Build/packaging config (the non-obvious decisions — and where CI failed)
  - projects/task-explorer/{examples/npm_task_contrib,examples/vscode_task_contrib,packages/tasks_panel}/scripts/package-vsix.js — `useYarn: false` + `dependencies: false`. These scripts were lifted from the yarn-based source repo with useYarn: true; under @vscode/vsce 2.x that runs yarn list, which reads the root packageManager: pnpm@11.1.1 and hard-fails (exit 232). The extensions are fully webpack-bundled, so vsce needs no dependency walk. This was the original CI failure. 
  - nyc.config.js — excludes tasks_panel/src/** from the merged 100% coverage gate. Unlike other packages (which instrument dist/src/**, already excluded), tasks_panel instruments src/** directly, so its intentionally-untested paths (logger, webSocketServer, a few UI callbacks) leaked into the global gate. It keeps its own 99% gate in its own CI. (If you'd prefer true global 100%, the alternative is 2 small tests for extension.ts:41 and abstract-webview-panel.ts:91-93 — flag it.)
  - projects/task-explorer/webpack.config.vscode.base.js — shared VSCode-extension webpack base; note resolve.modules points at the workspace root node_modules first (pnpm hoisting) and minimize: false (avoids a known runtime error + reduces packaged file count).
  - projects/task-explorer/tsconfig.base.json — extends root base, sets noImplicitAny: false + skipLibCheck: true to match the source repo's historical behaviour.

  3. Lint config (the non-obvious decisions)
  - .eslintrc.js — one projects/task-explorer/** override block relaxing the rules that fire on the migrated code (no-unsafe-*, no-explicit-any, no-unused-vars, etc.), plus a repo-wide ignorePatterns: ["**/dist/", "**/coverage/"]. TODO comments left to tighten incrementally — mirrors prior migrations.
  
  4. Version/lockfile alignment (small but load-bearing)
  - @types/vscode aligned to 1.75.0 across the examples (57f461e).
  - pnpm-lock.yaml vue-resolution fix (6f61c17) — the earlier merge of main had downgraded inquirer-gui's vue resolution while resolving lock conflicts; restored to unblock the tasks_panel build.

  5. Everything else — skim
  - src/, test/, Vue components, icons, README.md, CHANGELOG.old.md are a lift + mechanical hygiene. Not where risk lives.


jacob-kreyenbuehl (changes requested):
Hi Deeksha, nice work. I checked the migration against the source repo (SAP/task-explorer). Two reviewers checked it plus I verified the facts in the repo.

What looks good:
- All moved files match the source after Prettier. No source changes lost.
- Examples moved from `packages/` in the source repo. That is fine.
- The `TreeItem.label` change (`typeof label === "string" ? ...`) is a real bug fix, not random drift. Correct for the VS Code type `string | TreeItemLabel`.
- Pins are right: `workspace:*` inside the project, exact pin for `@vscode-logging/logger`.
- `vue_frontend_rpc` is private and ignored in changesets. Correct, it is not on npm.

Two things to fix before merge:

**1. (should fix) Coverage gate excludes the whole tasks_panel src folder**

`nyc.config.js` adds `projects/task-explorer/packages/tasks_panel/src/**` to the root 100% coverage exclusion list. That hides ~42 source files from the monorepo gate. The existing exclusions are single files with a comment each. The cf-tools migration fixed its gap with real tests instead.

My suggestion: exclude only the files that really cannot be covered, one line per file with a short comment why. Plus open a tracking issue to reach 100% later.

**2. (should fix) Root ESLint ignore change affects every package**

`.eslintrc.js` adds `ignorePatterns: ["**/dist/", "**/coverage/"]` at root level. That changes linting for the whole repo, not just task-explorer. Low risk today (no tracked dist or coverage sources), but this kind of repo-wide change should not ride along inside a migration PR.

My suggestion: move it into the task-explorer block, or make it a separate PR with its own explanation.

**3. (nit) The new `TreeItemLabel` branch has no test**

There is no test with an object label (`label?.label`). And after finding 1, these lines are also outside the coverage gate. A small unit test in multi-step-select / tasks-tree with an object label would close this.

The relaxed ESLint rules for task-explorer (no-explicit-any, unsafe-*, floating-promises, ...) are scoped to `projects/task-explorer/**`. That matches how cf-tools and feature-toggle-node did it. OK for the import, as long as the TODO to tighten later is tracked.

Happy to help with any of these. Thanks!

deekshas8:
Thanks for the review @jacob-kreyenbuehl . I addressed points 1 and 2.

> (nit) The new TreeItemLabel branch has no test

Investigated this a bit. Both the tasks-tree.ts and multi-step-select.ts object-label branches sit under /* istanbul ignore next */, so they're already excluded from the coverage gate. That means a unit test wouldn't actually move coverage. I can add one for the object-label case if we want the safety net, but wanted to flag that it won't change the gate number.

jacob-kreyenbuehl (approved):
Re-checked after `dc24112` ("narrow eslint ignore and coverage-gate exclusions"). Verified against the repo:

**Finding 1 (coverage gate): fixed.** The wholesale `tasks_panel/src/**` exclusion is gone. Now only:
- `src/logger/**` and `src/webSocketServer/**` (never instrumented by the package's own nyc — a faithful mirror)
- `src/extension.ts` and `src/panels/abstract-webview-panel.ts` (individually excluded with a comment noting they're below 100%)
Everything else is back under the monorepo 100% merged gate. The comment in `nyc.config.js` ("mirror the dirs the package's own nyc.config.js excludes") is accurate for the two dirs. (Small note: the 2 files are not in the package's own nyc exclude, so "mirror" only strictly applies to the two dirs — fine either way, the per-file comments cover it.)

**Finding 2 (eslint ignorePatterns): fixed.** Now scoped to `projects/task-explorer/**/dist/` and `projects/task-explorer/**/coverage/`. No more repo-wide glob.

**Finding 3 (nit, untested TreeItemLabel branch):** still open, but it is a nit. The branch is in `tasks-tree.ts` which is NOT in the narrowed exclusion, so it counts toward the gate. Not blocking.

Local verification done on `45b80e7`: eslint on projects/task-explorer/**/*.ts = clean; tasks_panel package ci = rc 0 (incl. vsix packaging); scripts/merge-coverage = rc 0 (the narrow exclusions hold up under the real merged gate). GitHub CI (Full Build, commitlint, CLA) = all pass on the current head.

LGTM. Approving.

PR #620, opened 2026-09-21
chore: migrate xml-tools

rolanbadrislamov:
## Summary

Migrates [`SAP/xml-tools`](https://github.com/SAP/xml-tools) into `projects/xml-tools/`

**What changed:**

- 10 packages moved into `projects/xml-tools/packages/` (9 published `@xml-tools/*` libraries + the private `xml-toolkit` VS Code extension)
- Wired into the monorepo toolchain: `workspace:*` intra-project deps, per-package `tsconfig`/`nyc` config, root eslint overrides
- Added transitive deps that yarn's flat `node_modules` had masked (`chevrotain`, `@xml-tools/ast`, `@xml-tools/common`) — surfaced by pnpm's strict isolation
- `language-server` now detects its bundled `dist/server.js` directly instead of probing `node_modules` (unreliable under pnpm symlinks)
- `xml-toolkit` VSIX packaging reworked for pnpm (was coupled to yarn `nohoist`); `.vscodeignore`/`.gitignore` tightened
- `@types/vscode` kept pinned at `1.56.0` (minimum supported API); rationale noted in `package.json`

## Reviewer's guide

Review in this order:

1. **`.eslintrc.js`** — new overrides scoped to `projects/xml-tools/**`, relaxing chevrotain-generated + `api.d.ts` violations; mirrors the vscode-logging pattern
2. **`projects/xml-tools/packages/*/package.json`** — no `private` on the libraries, `workspace:*` deps, `repository.directory`/`bugs` point at the monorepo, `files` includes CHANGELOG
3. **`projects/xml-tools/packages/xml-toolkit/scripts/package-vsix.js`** — vsce hot-patch resolves the sibling `language-server` via its `node_modules` symlink so it packages as a child; LICENSE copy is crash-safe (before pkg.json mutation)
4. **`projects/xml-tools/packages/language-server/lib/api.js`** — `SERVER_PATH` detection change (only non-move logic edit)
5. **Sources under `packages/*/lib` & `api.d.ts`** — source lift only, no logic changes

## Not in this PR

- **VSIX runtime check** — the extension is build-verified (VSIX packages with the bundled server); a manual install in a BAS dev-space to confirm XML features activate is a separate step


jacob-kreyenbuehl (commented):
Reviewed the migration end to end. Nice work, the source lift is clean and CI is green (Full Build compiles, tests, merges coverage, lints, and packages the VSIX).

## What I verified (all good)

- **Source fidelity**: I diffed every moved `lib`/`test`/`api.d.ts` against SAP/xml-tools. 242 files are byte-identical. Only 4 differ: 3 are pure prettier reflow, 1 is the documented `api.js` change.
- **`api.js` SERVER_PATH change**: correct. The old check looked at `node_modules` existence, which does not work under pnpm (pnpm always makes a symlinked `node_modules`). Detecting the bundled `dist/server.js` directly is the right fix. `bundledPath` matches the webpack output.
- **VSIX packaging** (`package-vsix.js`): well documented, and CI proves it really builds a VSIX. The sibling-via-symlink trick (so the package is a child of the extension dir) is sound.
- **Coverage**: the 9 libraries run `coverage:*` and hit the root 100% gate. `xml-toolkit` runs only `clean bundle package`, so it opts out. This matches the pattern we use for extension backends.
- **Pinning / privacy / deps**: no external consumers of `@xml-tools/*`, 21 `workspace:*` intra deps, only `xml-toolkit` is private, and the pnpm strict-isolation deps (chevrotain, ast, common) are declared. The `@types/vscode` pin has a clear reason comment.

## Findings

### 1. Should fix (owner call): the `fixed` changeset group changes the version policy and would push a fake major bump to public users

The source repo uses lerna `"version": "independent"`. The 9 packages are on purpose at different versions (ast 5.1.0, common 0.2.0, parser 1.0.11).

A `fixed` group forces lockstep. I checked this on the repo: both existing fixed groups (vscode-logging at 2.0.9, inquirer-gui at 3.4.13) came from sources that were **already unified** in lerna, so the fixed group kept their old behavior. xml-tools is the first one coming from an **independent** source, so here the fixed group is a real policy change.

What happens: the next changeset that touches any one package will snap **all 9 to one version**. So `@xml-tools/common` goes 0.2.0 to 5.1.x in a single release. These are public packages with heavy use (`@xml-tools/ast` is about 22k downloads/week), so thousands of external users get a fake 5 major jump, and every release republishes all 9 even when only 1 changed.

`updateInternalDependencies: patch` already bumps dependents for you, so lockstep is not needed for correctness. My rec: drop the fixed group and let them version independently, same as the source. If lockstep is what you want, it is a policy change on public packages, so please get an owner sign off first.

### 2. Nit: `["xml-toolkit"]` as a single-item fixed group does nothing

A fixed group needs 2 or more members. task-explorer's private extension is just left out of the config and versions independently by default. Your PR text also says "xml-toolkit versions independently", which is exactly what leaving it out gives you. You can remove this line.

### 3. Nit: dead script reference

`xml-toolkit` has `"vsix:prepare": "sh prepare-vsix-package.sh"`, but that file does not exist. Nothing in `ci` calls it, so it is harmless, but it is dead. Suggest removing it.

### 4. Needs verify (low): deprecated `vsce@1.84.0`

The monorepo standard is `@vscode/vsce@2.24.0`. It works here because the proxyquire hot patch depends on vsce 1.x internals (`vsce/out/package`, the `./npm` `getDependencies`). This is fine for now because xml-toolkit has to bundle the sibling language-server `dist/`, which is harder than plain `--no-dependencies`. But it is a deprecated package with old transitive deps. Not a blocker, just worth a note for later cleanup.

Overall this is a solid, faithful migration. Finding 1 is the only real one, and it is a versioning policy question for public packages, so it needs an owner decision.

rolanbadrislamov:
> Reviewed the migration end to end. Nice work, the source lift is clean and CI is green (Full Build compiles, tests, merges coverage, lints, and packages the VSIX).

FIndings 1, 2, 3 are addressed in [d601cc3](https://github.com/SAP/app-studio-toolkit/pull/620/commits/d601cc35ef2e893391faab93195dc20630f5c291)

jacob-kreyenbuehl (approved):
Confirmed, both findings are resolved correctly.

- Dropping the fixed group puts the `@xml-tools/*` packages back on independent versioning, same as the source repo. Verified the config is valid and xml-tools is now in neither `fixed` nor `ignore`.
- `["xml-toolkit"]` single-item group removed, and the dead `vsix:prepare` script is gone with no hanging references.
- CI re-ran on the new commit and is fully green.

Thanks for addressing these cleanly. Approving.

